PR #5, opened 2021-08-19
example to test new save data method from socket

guinneak198:
I tried to simplify the run_fieldSweep.py code to give a minimal example to debug the saving. If I left any unneccessary lines in please do let me know!


examples/save_data_example.py
```diff
@@ -0,0 +1,92 @@
...
+    script, as this function is not designed to be moved outside the module
+    """
+    for x in range(nScans):
+        run_scans_time_list = [time.time()]
+        SpinCore_pp.configureTX(26, 14.894179, r_[0.0,90.0,180.0,270.0], 1.0, 2048)#adc_offset, carrierFreq_MHz, tx_phases, amplitude, nPoints
+        acq_time_ms = SpinCore_pp.configureRX(24, 2048, 1, 1, 4)#SW_kHz, nPoints, 1, nEchoes, nPhaseSteps
+        SpinCore_pp.init_ppg()
+        SpinCore_pp.load([
+            ('marker','start',1),
+            ('phase_reset',1),
```
jmfranck, line 33: so I don't understand why you don't just delete the entire run_scans function and replace it with a delay (where it's called below).  Nothing inside run_scans affects the functioning of your log, and then you have removed the dependence on SpinCore_pp in one swoop.

examples/save_data_example.py
```diff
@@ -0,0 +1,92 @@
...
+        if j == 0:
+            MWfreq = p.dip_lock(9.81,9.83)
+            p.start_log()
+        p.set_power(this_dB)
+        for k in range(10):
+            time.sleep(0.5)
+            if p.get_power_setting() >= this_dB: break
+        if p.get_power_setting() < this_dB: raise ValueError("After 10 tries, the power has still not settled")
+        time.sleep(5)
+        meter_powers[j] = p.get_power_setting()
```
jmfranck, line 76: I think we agreed that there should be no dip locking or setting of the power.

jmfranck: on the other hand, you do want to be sure that you keep the command that starts the log!!

jmfranck: actually, you should be able to leave these in -- [see slack](https://jmfrancklab.slack.com/archives/CV553TPRV/p1629401557000900?thread_ts=1629396467.000300&cid=CV553TPRV)

examples/save_data_example.py
```diff
@@ -0,0 +1,92 @@
...
+        with xepr() as x_server:
+            for B0_index,desired_B0 in enumerate(r_[3503:3508:0.1]):
+                true_B0 = x_server.set_field(desired_B0)
```
jmfranck, line 79: I would also strip out any references to the `x_server`, since we don't believe it's the source of your problems.

examples/save_data_example.py
```diff
@@ -0,0 +1,92 @@
...
+                print("My field in G is %f"%true_B0)
+                time.sleep(3.0)
+                carrierFreq_MHz = 0.00425*true_B0
+                carrierFreqs_MHz[B0_index] = carrierFreq_MHz
+                fields_Set[B0_index] = true_B0
+                print("My frequency in MHz is",carrierFreq_MHz)
```
jmfranck, line 85: I think that this printing and math is also tangential to your issue

examples/save_data_example.py
guinneak198: deleted the run_scans function

guinneak198: okay replaced with delay

examples/save_data_example.py
guinneak198: okay took out references to xepr_server and spincore

examples/save_data_example.py
```diff
@@ -6,8 +12,8 @@
 import sys
 import time
 import random
-long_delay = time.sleep(5)
-short_delay = time.sleep(1)
+long_delay = 1e-3
+short_delay = 1e-4
```
jmfranck: previous code what not doing what you think, most likely.

examples/save_data_example.py
```diff
@@ -65,28 +71,28 @@ def run_scans(nScans, power_idx, field_idx, DNP_data=None):
 
 with power_control() as p:
     for j,this_dB in enumerate(dB_settings):
+        print("I'm going to pretend to run",this_dB,"dBm")
         if j == 0:
-            short_delay
+            time.sleep(short_delay)
```
jmfranck: paired w/ edit above.

examples/save_data_example.py
```diff
@@ -65,28 +71,28 @@ def run_scans(nScans, power_idx, field_idx, DNP_data=None):
...
             if p.get_power_setting() >= this_dB: break
-        long_delay
+        time.sleep(long_delay)
         meter_powers[j] = p.get_power_setting()
         if True:
+            # the following seems unrealistic for a field sweep -- what's
+            # up with that?
             for B0_index,desired_B0 in enumerate(r_[3501:3530:0.001]):
```
jmfranck: what's up here?  I can't believe you were really trying to run an experiment where you used 29,000 field sweep steps.

examples/save_data_example.py
```diff
@@ -65,28 +71,28 @@ def run_scans(nScans, power_idx, field_idx, DNP_data=None):
...
                 #carrierFreq_MHz = rand()
                 carrierFreqs_MHz[B0_index] = rand()
                 fields_Set[B0_index] = rand()
-                short_delay
+                time.sleep(short_delay)
                 if True:
-                    if DNP_data is None:
-                        DNP_data = run_scans(1,j,B0_index, DNP_data = None)
-                    else:
-                        DNP_data = run_scans(1,j,B0_index,DNP_data)
```
jmfranck: this was redundant (if `DNP_data` is None, why explicitly pass it as `None`?)

jmfranck:
`test_error.py` doesn't run -- it seems like you are passing a multi-dimensional array to the hermitian function, and that it doesn't like this?  We should verify that that's what's up and what we want to do about it.

examples/save_data_example.py
guinneak198: I am exaggerating the length to show that for HUGE amounts it will still work

PR #8, opened 2021-11-05
Combined ODNP with server-based log

jmfranck:


examples/just_enhancement.py
```diff
@@ -120,7 +120,6 @@ def run_scans(nScans, power_idx, DNP_data=None):
         print("\nRUNNING BOARD...\n")
         this_time_var = time.time()
         SpinCore_pp.runBoard()
-        power_start_times.append(this_time_var)
```
jmfranck: I didn't like the scope of this -- list is created outside the function, at the top, but here you're manipulating it like a global variable.

examples/just_enhancement.py
```diff
@@ -150,6 +149,7 @@ def run_scans(nScans, power_idx, DNP_data=None):
         return DNP_data
 #}}}
 ini_time = time.time() # needed b/c data object doesn't exist yet
+power_start_times.append(time.time())
 DNP_data = run_scans(nScans,0)
```
jmfranck: this is much more clear, anyways.

examples/just_enhancement.py
```diff
@@ -165,6 +165,7 @@ def run_scans(nScans, power_idx, DNP_data=None):
         p.set_power(this_dB)
         time.sleep(5)
         power_settings[j] = p.get_power_setting()
+        power_start_times.append(time.time())
         run_scans(nScans,j+1,DNP_data)
```
jmfranck: This is much more clear, anyways.

examples/just_enhancement.py
jmfranck: and here we store the time for the thermal scan.

examples/just_enhancement.py
```diff
@@ -173,15 +174,6 @@ def run_scans(nScans, power_idx, DNP_data=None):
...
 # can delete this after
-print("*** *** *** *** *** *** ***")
-print("*** *** *** *** *** *** ***")
-print("*** *** *** *** *** *** ***")
-print("THIS IS MY POWER START TIMES",power_start_times)
-print("*** *** *** *** *** *** ***")
-print("*** *** *** *** *** *** ***")
-print("*** *** *** *** *** *** ***")
-for q in range(len(power_start_times)):
-    power_start_times[q] = power_start_times[q] - power_start_times[0]
```
jmfranck: you don't want to do this -- just store the absolute time in the data.  You can make it relative during processing.

examples/just_enhancement.py
jmfranck: I will add that I'm a bit confused, because my power_start_times would be 1 longer than yours --> I actually don't understand how the length of yours matches the length of your indirect dimension, and/or where the 0 at the beginning comes from.   So, definitely, you would want to test + debug this with 3 powers before trying a full run.

proc_fieldSweep.py
```diff
@@ -0,0 +1,61 @@
+from pylab import *
```
guinneak198, line 1: not sure why this is in this PR, I tried diffing against master but nothing came up so I think that means there is no difference but then why is it showing up here?

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
+import os, time, h5py
```
guinneak198, line 1: Warning: this data is bad due to either bad tuning or something. However, the matching of the time/power log with the enhancement and IR to get the averaged power works so this PR is primarily focused on that goal

guinneak198 (commented):
PR ready in the sense it takes the time logs from both IR and enhancement and matches with the averaged power over those times to apply. thermal scans are hard set to 0. Be aware the data collected is awful and so the processing part is rough. But the power axis log looks good

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -84,7 +84,7 @@ def verifyParams():
 # as this is generally what the SpinCore takes
 # note that acq_time is always milliseconds
 #}}}
-p90 = 4.69
+p90 = 1.781
```
jmfranck, line 87: I notice that this is massively different -- what's up with that?

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
...
-            meter_powers = zeros_like(dB_settings)
-            for j,this_power in enumerate(r_[dB_settings]):
-                print("\n*** *** *** *** ***\n") 
-                if j>0 and this_power > last_power + 3:
+    with Bridge12() as b:
+        b.set_wg(True)
+        b.set_rf(True)
+        b.set_amp(True)
+        this_return = b.lock_on_dip(ini_range=(9.819e9,9.825e9))
+        dip_f = this_return[2]
```
jmfranck, line 126: the way of  doing that is `_, dip_f = b.lock...`

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
...
+                        next_power += 3
+                b.set_power(this_power)
+            else:
+                b.set_power(this_power)
+            time.sleep(15)
+            with prologix_connection() as p:
+                with gigatronics(prologix_instance=p, address=7) as g:
+                    meter_powers = g.read_power()
+                    print("POWER READING",meter_powers)
+        for B0_index,desired_B0 in enumerate(field_axis):
```
jmfranck, line 163: OK -- so here is your field sweep -- you moved it inside the B12 setup, which is good.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,352 @@
...
+        nScans=nScans, power_idx=0)
+time_list.append(time.time())
+time_axis_coords_IR = vd_data.getaxis('indirect')
+time_axis_coords_IR[0]=ini_time
+vd_data.set_prop('start_time',ini_time)
+meter_power=0
+save_file = True
+acq_params = {j:eval(j) for j in dir() if j in ['adcOffset', 'carrierFreq_MHz', 'amplitude',
+    'nScans', 'nEchoes', 'p90_us', 'deadtime_us', 'repetition_us', 'SW_kHz',
+    'nPoints', 'tau_adjust_us', 'deblank_us', 'tau_us', 'MWfreq', 'acq_time_ms', 'meter_power']}
```
jmfranck, line 250: this is the new method for retrieving acq_params from the global namespace

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
...
+            true_B0 = x_server.set_field(desired_B0)
+            print("My field in G is %f"%true_B0)
+            time.sleep(3.0)
+            carrierFreq_MHz = gamma_eff*true_B0
+            print("My frequency in MHz is",carrierFreq_MHz)
+            acq_params['carrierFreq_MHz'] = carrierFreq_MHz
```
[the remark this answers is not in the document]
jmfranck: OK -- what I say there isn't using quite the right words -- rather, you want to pull these variables automatically from the global namespace -- that's done [here](https://github.com/jmfrancklab/spincore_apps/pull/8/files#r769073960)

in general, you should not be messing directly with `acq_params['...']`, since the code I link to there does it for you.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,352 @@
...
+        print("CONFIGURING RECEIVER...")
+        acq_time_ms = SpinCore_pp.configureRX(SW_kHz, nPoints, 1, nEchoes, nPhaseSteps)
+        # acq_time_ms is in msec!
+        print("ACQUISITION TIME IS",acq_time_ms,"ms")
+        verifyParams(nPoints=nPoints, nScans=nScans, p90_us=p90_us, tau_us=tau_us)
+        run_scans_time_list.append(time.time())
+        run_scans_names.append('init')
+        print("\nRECEIVER CONFIGURED.")
+        print("***")
+        print("\nINITIALIZING PROG BOARD...\n")
```
jmfranck, line 98: do you actually read and use this?  scan number might be useful, but everything else that's noise should go to the log file, I would think.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,352 @@
...
+        print("SAVING FILE...")
+        DNP_data.hdf5_write(myfilename)        
+        print("FILE SAVED")
+        print("Name of saved enhancement data", DNP_data.name())
+        print("shape of saved enhancement data", ndshape(DNP_data))
+        save_file=False
+    except Exception as e:
+        print(e)
+        print("EXCEPTION ERROR - FILE MAY ALREADY EXIST.")
+        save_file = False
```
jmfranck, line 348: here, set save_file to False.  Add `time.sleep(3)`.  Then, it would keep trying until the "try" part succeeded or you pressed ctrl-C or killed the process.

The point would be that you could rename the old file in a separate window -- then it could save with the file name you had given, and you would not lose your data.

You would want to edit the EXCEPTION ERROR message here to explain what's going on, though, because you are deliberately setting up an infinite loop.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+import pylab as plt
+from numpy import empty
+from sympy import exp as s_exp
+from sympy import symbols, Symbol
+from matplotlib.ticker import FuncFormatter
+import matplotlib.transforms as transforms
+from pyspecdata import *
+from pyspecProcScripts import *
+from pyspecProcScripts import lookup_table
```
jmfranck, line 10: the point of this is just to check the data -- so while this might form the basis of a larger code on procScripts, you don't want all the bells and whistles.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+from Instruments.logobj import logobj
+logger = init_logging("info")
+Ep_signal_pathway = {'ph1':1}
+fl=fl_mod()
+filename='211213_10mM_TEMPOL_test'
+file_location = 'ODNP_NMR_comp/ODNP'
+Ep_f_slice=(0.15e3,0.4e3)
+IR_f_slice=(0.15e3,0.4e3)
+Ep_t_range = (0,0.2)
+t_range = (0,0.2)
```
jmfranck, line 20: since you're hard-coding all these things (fine for code where you're just checking the data), then also include the time for the center of the echo -- so that you don't have to do any automated phasing.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+excluded_pathways = [(0,0)]
+R1w = 1/2.172
+nPowers = 15
+ppt = 1.5167e-3
+C=0.01
+#{{{load in log
+with h5py.File(search_filename(filename+".h5",exp_type='ODNP_NMR_comp/ODNP',unique=True),'r') as f:
+    log_grp = f['log']
+    thislog = logobj()
+    thislog.__setstate__(log_grp)
```
jmfranck, line 30: a little more advanced, but I think we can write a [classmethod](https://realpython.com/instance-class-and-static-methods-demystified/) to compress these steps

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    read_array = thislog.total_log
+    read_dict = thislog.log_dict
+for j in range(len(read_array)):
+    thistime,thisrx,thispower,thiscmd = read_array[j]
```
jmfranck, line 34: you're looping over an array, which is never good -- what is `read_array` -- is it a structured array?

jmfranck: you can answer this question by looking at `read_array.dtype`.  If it is a structured array, what are the fields?  I can give you the code to replace this.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+fig, (ax_Rx,ax_power) = plt.subplots(2,1, figsize=(10,8))
+fl.next("log figure",fig=fig)
```
jmfranck, line 36: I don't think you're benefitting from the figure list here -- I think just use matplotlib with no figure list -- this is followed w/ some other comments.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    ax_power.axvline(x=thisevent['time']-log_start_time)
+    y_pos = j/n_events
+    #ax_Rx.text(thisevent['time']-log_start_time,y_pos,read_dict[thisevent['cmd']],transform=trans_Rx)
+    #ax_power.text(thisevent['time']-log_start_time,y_pos,read_dict[thisevent['cmd']],transform=trans_power)
+ax_power.legend(**dict(bbox_to_anchor=(1.05,1),loc=2,borderaxespad=0.))
+plt.tight_layout()
+power_axis = nddata(read_array['power'],[-1],['time'])
+power_axis.setaxis('time',relative_time)
+power_axis.setaxis('time',lambda x: x - log_start_time)
+fl.next('power axis')
```
jmfranck, line 61: following on comment above -- this can be replaced with `figure(); title('power axis')`

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+fl.plot(power_axis,'.')
+power_axis = nddata(read_array['power'],[-1],['time'])
+power_axis.setaxis('time',relative_time)
+power_axis.name('power')
+fl.next('power axis W')
```
jmfranck, line 66: analogous to prev.  comment

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+trans_power = transforms.blended_transform_factory(
+        ax_power.transData,ax_power.transAxes)
+trans_Rx = transforms.blended_transform_factory(
+        ax_Rx.transData, ax_Rx.transAxes)
+for j, thisevent in enumerate(read_array[mask]):
+    ax_Rx.axvline(x=thisevent['time']-log_start_time)
+    ax_power.axvline(x=thisevent['time']-log_start_time)
+    y_pos = j/n_events
+    #ax_Rx.text(thisevent['time']-log_start_time,y_pos,read_dict[thisevent['cmd']],transform=trans_Rx)
+    #ax_power.text(thisevent['time']-log_start_time,y_pos,read_dict[thisevent['cmd']],transform=trans_power)
```
jmfranck, line 55: if you're not using the text, then you don't need the transforms above.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+ax_power.legend(**dict(bbox_to_anchor=(1.05,1),loc=2,borderaxespad=0.))
+plt.tight_layout()
+power_axis = nddata(read_array['power'],[-1],['time'])
+power_axis.setaxis('time',relative_time)
+power_axis.setaxis('time',lambda x: x - log_start_time)
+fl.next('power axis')
+fl.plot(power_axis,'.')
+power_axis = nddata(read_array['power'],[-1],['time'])
+power_axis.setaxis('time',relative_time)
```
jmfranck, line 64: I don't understand why this is done a second time?

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+power_axis.name('power')
+fl.next('power axis W')
+coupler_atten = 22
+power_axis.data = (10**((power_axis.data+coupler_atten)/10+3))/1e6 #convert to W
```
jmfranck, line 68: formula is weird -- where does 1e6 come from?

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+start_times = []
+stop_times = []
+errors=[]
+for nodename, postproc, signal_pathway, clock_correction in [
+        ('FIR_noPower','spincore_IR_v1',{'ph1':0,'ph2':1},False),
+        ('FIR_27dBm','spincore_IR_v1',{'ph1':0,'ph2':1},False),
+        ('FIR_30dBm','spincore_IR_v1',{'ph1':0,'ph2':1},False),
+        ('FIR_32dBm','spincore_IR_v1',{'ph1':0,'ph2':1},False),
+        ('FIR_33dBm','spincore_IR_v1',{'ph1':0,'ph2':1},False),
+        ]:
```
jmfranck, line 83: signal pathway outside -- b/c this accompanies the pulse program, you know they all have the same  pathway.  Remove clock correction option to simplify (see general comment up front)

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    IR = find_file(filename,exp_type=file_location,expno=nodename,
+            postproc=postproc,lookup=lookup_table)
```
jmfranck, line 85: big problem here -- postproc argument should not be used or required -- your ppg should be setting the appropriate attribute of the nddata so that it knows which postproc to use.  and the name should NOT be version 1, since you are using the log, etc, etc, which is not part of version 1!

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    times = IR.C.getaxis('indirect').copy()
+    IR.mean('indirect')
```
jmfranck, line 87: why are you taking the mean along the indirect dimension??

jmfranck: what is the indirect dimension here?

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    fl.next('Raw IR')
+    fl.image(IR.C.setaxis('vd','#').set_units('vd','scan #'))
+    IR['ph2',0]['ph1',0]['t2':0]=0 #kill axial noise
+    IR.ift('t2')
+    IR.ift(['ph1','ph2'])
+    t_start = t_range[-1]/4
+    t_start *= 3
+    rx_offset_corr = IR['t2':(t_start,None)]
+    rx_offset_corr = rx_offset_corr.data.mean()
+    IR -= rx_offset_corr
```
jmfranck, line 97: doesn't seem important for a basic check of the signal quality.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    IR.ft('t2')
+    IR.ft(['ph1','ph2'])
+    zero_crossing = abs(select_pathway(IR['t2':IR_f_slice],signal_pathway)).C.sum('t2').argmin('vd',raw_index=True).item()
+    IR=IR['t2':IR_f_slice]
```
jmfranck, line 101: extra junk

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+            fl.next('clock correction')
+            fl.plot(s_clock,'.',alpha=0.7)
+        clock_corr = s_clock.argmax('clock_corr').item()
+        plt.axvline(x=clock_corr, alpha=0.5, color='r')
+        IR *= np.exp(-1j*clock_corr*IR.fromaxis('vd'))
+        IR.ft(['ph1','ph2'])
+        if fl is not None:
+            fl.next('after auto-clock correction')
+            fl.image(IR.C.setaxis('vd','#'))
+        IR.ift('t2')
```
jmfranck, line 128: not using clock correction + not important for basic check of signal quality

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    #}}}
+    #{{{phasing
+    IR.ft('t2')
+    IR.ift('t2')
+    best_shift = hermitian_function_test(select_pathway(IR.C.mean('vd'),
+        signal_pathway),aliasing_slop=0)
+    best_shift = 0.0035 #this data is bad (bad tuning) so I hard set the shift
```
jmfranck, line 135: you're hard-coding the best shift here, anyways -- which is what I'm suggesting at the top
so, remove the dependence on the phasing routine.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    IR.setaxis('t2',lambda x: x-best_shift).register_axis({'t2':0})
+    IR.ft('t2')
+    IR.ift('t2')
```
jmfranck, line 138: just do this at the very beginning.  we don't need to see the multiple steps here.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    IR.ift(['ph1','ph2'])
+    opt_shift,sigma, my_mask = correl_align(IR.C*mysgn,indirect_dim='vd',
+            signal_pathway=signal_pathway,sigma=150)#again, bad data
+    IR.ift('t2')
+    IR *= np.exp(-1j*2*pi*opt_shift*IR.fromaxis('t2'))
+    IR.ft('t2')
+    IR.ift('t2')
+    IR.ft(['ph1','ph2'])
+    IR.ft('t2')
+    #}}}
```
jmfranck, line 153: no alignment

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    d.ft('t2')
+    #}}}
+    #{{{Integrate with error
+    error_pathway = (set(((j,k) for j in range(ndshape(d)['ph1']) for k in range(ndshape(d)['ph2'])))
+            - set(excluded_pathways)
+            -set([(signal_pathway['ph1'],signal_pathway['ph2'])]))
+    error_pathway = [{'ph1':j,'ph2':k} for j,k in error_pathway]
+    s_int,frq_slice = integral_w_errors(d,signal_pathway,error_pathway,
+            convolve_method='Lorentzian',
+            indirect='vd',return_frq_slice=True)
```
jmfranck, line 168: this is good to keep

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    #{{{phasing
+    IR.ft('t2')
+    IR.ift('t2')
+    best_shift = hermitian_function_test(select_pathway(IR.C.mean('vd'),
+        signal_pathway),aliasing_slop=0)
+    best_shift = 0.0035 #this data is bad (bad tuning) so I hard set the shift
+    IR.setaxis('t2',lambda x: x-best_shift).register_axis({'t2':0})
+    IR.ft('t2')
+    IR.ift('t2')
+    IR /= zeroth_order_ph(select_pathway(IR,signal_pathway))
```
jmfranck, line 139: it would be harder to replace this with something manual, so leave it

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    x = s_int.fromaxis('vd')
+    M0,Mi,R1,vd = symbols("M_0 M_inf R_1 vd",Real=True)
+    functional_form = Mi + (M0-Mi) * s_exp(-vd*R1)
+    f = fitdata(s_int)
+    f.functional_form = functional_form
+    f.fit()
+    T1 = 1./f.output('R_1')
+    fit = f.eval(100)
+    T1_list.append(T1)
+    #}}}
```
jmfranck, line 180: don't fit.

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+        nddata_p_vs_t.set_error(errors)
+        fl.plot(nddata_p_vs_t,'ro',capsize=6)
+logger.info(strm("T1 list:",T1_list)) #we don't expect good T1s since bad tuning
+logger.info(strm("Power list:",power_list))
+#}}}
+#{{{Load process enhancement
+for filename,nodename,file_location in [
+        (filename,'enhancement_curve','ODNP_NMR_comp/ODNP')
+        ]:
+    #{{{
```
jmfranck, line 210: please review this and below following general themes from above.

examples/run_nutation.py
```diff
@@ -27,7 +27,7 @@ def verifyParams():
         quit()
     else:
         print("VERIFIED PULSE TIME.")
-    if (tau < 0.065):
+    if (tau_us < 0.065):
```
jmfranck, line 30: name change can't be right -- this would mean 65 ns

examples/run_nutation.py
```diff
@@ -61,84 +61,65 @@ def API_sender(value):
...
-acq_time = nPoints/SW_kHz # ms
-tau_adjust = 0.0
-tau = 1000
-print("ACQUISITION TIME:",acq_time,"ms")
-print("TAU DELAY:",tau,"us")
+acq_time_ms = nPoints/SW_kHz # ms
+tau_adjust_us = 0.0
+tau_us = deadtime_us + acq_time_ms*1e3*0.5 + tau_adjust_us
+print("ACQUISITION TIME:",acq_time_ms,"ms")
+print("TAU DELAY:",tau_us,"us")
```
jmfranck, line 84: these seem to be parameter changes -- use the same parameters (i.e. hard-coded settings) as master.

jmfranck: (we don't want irrelevant changes in git)

jmfranck: line 82 seems to be relevant

examples/struct_array_example.py
```diff
@@ -0,0 +1,22 @@
...
+mytimes[:]['stop_time'] = r_[0:10] + 1
+mydata = nddata(r_[0:10]*10,[-1],['t'])
+# I can attach my structured array just like any other normal axis
+mydata.setaxis('t', mytimes)
+mydata.name('data')
+mydata.hdf5_write('struct_array_axis_example.h5')
+mydata_read = nddata_hdf5('struct_array_axis_example.h5/data')
+# now access the start vs. stop times
+print("start times:",mydata_read.getaxis('t')['start_time'])
+print("stop times:",mydata_read.getaxis('t')['stop_time'])
```
jmfranck, line 22: move this into a subdirectory or somesuch, so it doesn't show up in the gallery.
btw, I haven't tried re-running the gallery -- you should do that too

proc_fieldSweep.py
jmfranck: this is not in master

jmfranck: this should not be added to the root directory. move to examples?  this does seem very minimalist, which is good.

examples/proc_combined_ODNP.py
guinneak198: the hard coded time for the center of the echo should be the tau value right?

examples/proc_combined_ODNP.py
guinneak198: okay then for now until that classmethod is made I'll keep this here. Should that be a separate PR/new wrike task?

examples/proc_combined_ODNP.py
guinneak198: I took this from the test_power_control_server_read.py example in inst_notebooks. After printing read_array it is an ndarray and looks like a structured array 
`[(1.63960294e+09, -1. , -10.095 ,  8060245099936175237)
 (1.63960294e+09, 41.8,   7.3434, -8640926761156270880)
 (1.63960294e+09, 42.4,   7.3496, -8640926761156270880) ...
 (1.63960717e+09, 98.2,   6.2436,                    0)
 (1.63960717e+09, 98.6,   6.2437,                    0)
 (1.63960717e+09, 98.6,   6.2436, -7387592744514551751)]`

guinneak198: It didn't print the field names though so IDK as though there are any?

examples/proc_combined_ODNP.py
guinneak198: agreed

guinneak198: I should keep the axvlines though?

examples/proc_combined_ODNP.py
guinneak198: whoops sorry repeat!

examples/proc_combined_ODNP.py
guinneak198: So I know this works and produces the reasonable W. I originally got this from Alec's processing and after asking him he said it was too long ago to remember why the 1e6 is there. What would you suggest?

examples/proc_combined_ODNP.py
guinneak198: indirect dimension is the time recordings for when we compare with the log to find the average power over that time frame

examples/proc_combined_ODNP.py
guinneak198: version 1 is literally just reordering and fting the time and phase cycling dimensions in order to display the raw data. Why is that not okay for this processing?

guinneak198: but I edited the ppg to store the postproc for IR as an acq_params called IR_postproc...will this new method that I have load it correctly? I don't have data where it has this parameter since I JUST added it

examples/proc_combined_ODNP.py
guinneak198: frequency filtering is extra junk? the zero crossing is required especially for the higher powers where the signal tends to sometimes flip

examples/proc_combined_ODNP.py
```diff
@@ -0,0 +1,332 @@
...
+    error_pathway = (set(((j,k) for j in range(ndshape(d)['ph1']) for k in range(ndshape(d)['ph2'])))
+            - set(excluded_pathways)
+            -set([(signal_pathway['ph1'],signal_pathway['ph2'])]))
+    error_pathway = [{'ph1':j,'ph2':k} for j,k in error_pathway]
+    s_int,frq_slice = integral_w_errors(d,signal_pathway,error_pathway,
+            convolve_method='Lorentzian',
+            indirect='vd',return_frq_slice=True)
+    #}}}
+    #{{{Fitting Routine
+    x = s_int.fromaxis('vd')
```
guinneak198, line 118: do we still want to fit to acquire the T1?

examples/struct_array_example.py
guinneak198: spincore_apps doesn't have a doc_src folder? how/where do I run make html to test?

examples/proc_combined_ODNP.py
jmfranck: yes

examples/proc_combined_ODNP.py
jmfranck: yes -- as I say above, you should print `read_array.dtype` to get the field names, then, to answer this question.

examples/proc_combined_ODNP.py
jmfranck: yes

examples/proc_combined_ODNP.py
jmfranck: oh -- it's a hack that "just works."   get rid of the 1e6 and change the +3 to -3; then it makes sense, too!

examples/proc_combined_ODNP.py
jmfranck: what is the name of the property that you are setting now, and what's it set to? (so that I can check against pyspecdata)

examples/proc_combined_ODNP.py
jmfranck: I don't understand the answer -- is IR not NMR data?  Why are you averaging over a dimension of your NMR data?

examples/proc_combined_ODNP.py
jmfranck: no, not the frequency filtering -- I highlighted too much.

But the zero crossing, and switching back and forth between phase cycling and coherence domain -- yes.

You don't need the zero crossing.

examples/proc_combined_ODNP.py
jmfranck: have you added a singleton dimension, maybe?

examples/struct_array_example.py
jmfranck: never mind -- that is beyond the scope of the PR

SpinCore_pp/run_field_dep_justMw.py
jmfranck: OK that's fine

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
 #}}}
 
 with xepr() as x_server:
-    for B0_index,desired_B0 in enumerate(field_axis):
-        true_B0 = x_server.set_field(desired_B0)
```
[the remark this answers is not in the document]
jmfranck: it's fine as is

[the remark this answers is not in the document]
jmfranck: when I made this comment, I hadn't yet seen the other loop

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
...
-            dip_f = this_return[2]
-            print("Frequency",dip_f)
-            mw_freqs.append(dip_f)
-            acq_params['mw_freqs'] = mw_freqs
-            b.set_freq(dip_f)
-            meter_powers = zeros_like(dB_settings)
-            for j,this_power in enumerate(r_[dB_settings]):
-                print("\n*** *** *** *** ***\n") 
-                if j>0 and this_power > last_power + 3:
+    with Bridge12() as b:
```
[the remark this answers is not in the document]
jmfranck: I realized -- why are you doing this?? Shouldn't you be using the power control server instead?

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
...
+        b.set_rf(True)
+        b.set_amp(True)
+        this_return = b.lock_on_dip(ini_range=(9.819e9,9.825e9))
+        dip_f = this_return[2]
+        print("Frequency",dip_f)
+        mw_freqs.append(dip_f)
+        acq_params['mw_freqs'] = mw_freqs
+        b.set_freq(dip_f)
+        meter_powers = zeros_like(dB_settings)
+        for j,this_power in enumerate(r_[dB_settings]):
```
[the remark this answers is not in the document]
jmfranck: no -- add a docstring for the file that explains that it runs a field sweep, optionally at multiple microwave powers.

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -48,7 +48,7 @@ def verifyParams():
 #stop_field = 33510
 #field_axis = start_field + (stop_field-start_field)*uneven
 
-field_axis = r_[3502:3508:.5]
+field_axis = r_[3422:3426:.5]
```
jmfranck, line 51: don't commit trivial changes

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -63,10 +63,10 @@ def verifyParams():
 powers = 1e-3*10**(dB_settings/10.)
 #}}}
 
-output_name = '150uM_TEMPO_toluene_cap_probe_field_dep'
+output_name = 'TEMPOL_heat_exch_289uM_field_dep'
 node_name = 'field_sweep_1'
-adcOffset = 25
-gamma_eff = (14.892825/3505.43)
+adcOffset = 28
+gamma_eff = (14.549013/3424.42)
```
jmfranck, line 69: don't commit trivial changes (make them match master)

SpinCore_pp/run_field_dep_justMw.py
jmfranck: remove the trivial changes

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -94,7 +94,7 @@ def verifyParams():
 acq_time = nPoints/SW_kHz # ms
 tau_adjust = 0
 deblank = 1.0
-tau = 1000.
+tau = 3500.
```
jmfranck, line 97: don't commit trivial changes

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
...
-            SpinCore_pp.stopBoard();
-            print("FINISHED B0 INDEX %d..."%B0_index)
-            print("\n*** *** ***\n")
+                        next_power += 3
+                b.set_power(this_power)
+            else:
+                b.set_power(this_power)
+            time.sleep(15)
+            with prologix_connection() as p:
+                with gigatronics(prologix_instance=p, address=7) as g:
```
jmfranck, line 160: again, this should be handled by the power control server

SpinCore_pp/run_field_dep_justMw.py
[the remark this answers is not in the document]
jmfranck: the code hasn't changed here.

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -118,125 +118,125 @@ def verifyParams():
...
+                    data.setaxis('nScans',r_[0:nScans])
+                    data.setaxis('Field',field_axis)
+                    data.setaxis('power',r_[powers])
+                    data.name(node_name)
+                    data.set_prop('acq_params',acq_params)
+                data['nScans',x]['Field',B0_index]['power',0] = data_array
+                print("FINISHED B0 INDEX %d..."%B0_index)
+                print("\n*** *** ***\n")
+        last_power = this_power
+        SpinCore_pp.stopBoard();
```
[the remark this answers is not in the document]
jmfranck: the code hasn't changed.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+from SpinCore_pp.power_helper import gen_powerlist
+from pyspecdata.file_saving.hdf_save_dict_to_group import hdf_save_dict_to_group
+import h5py
+# {{{ from run_Hahn_echo_mw.py
+fl = figlist_var()
+save_file=True
+# {{{ experimental parameters
+# {{{ these need to change for each sample
+output_name = '150uM_TEMPOL'
+IR_postproc = 'spincore_IR_v1'
```
jmfranck, line 19: this should be called `postproc_type` -- pyspecdata looks for a property called `postproc_type`

examples/proc_combined_ODNP.py
jmfranck: see [here](https://github.com/jmfrancklab/spincore_apps/pull/8/files#r771069613)

examples/combined_ODNP.py
jmfranck: see [pyspecdata docs](https://jmfrancklab.github.io/pyspecdata/modules.html#pyspecdata.load_files.find_file)

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+acq_time_ms = 500. # ms
+nPoints = int(acq_time_ms*SW_kHz+0.5)
+tau_adjust_us = 0.0
+deblank_us = 1.0
+tau_us = 3500
+pad_us = 0
+pul_prog = 'ODNP_v3'
+vd_list = r_[2.1e3,2.1e4,2.1e5,4.3e5,6.4e5,8.8e5,1.1e6,1.3e6,1.5e6,1.7e6,1.9e6,2.5e6,3e6,3.3e6]
+T1_powers_dB = r_[30,32,33,34]
+T1_node_names = ['FIR_30dBm','FIR_32dBm','FIR_33dBm','FIR_34dBm']
```
jmfranck, line 43: `T1_node_names = ['FIR_%ddBm'%j for j in T1_powers_dB]`

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+nScans = 1
+nEchoes = 1
+date = datetime.now().strftime('%y%m%d')
+# all times in microseconds
+# note that acq_time_ms is always milliseconds
+p90_us = 4.464
+deadtime_us = 10.0
+repetition_us = 12e6
+FIR_rd = 7e6
+SW_kHz = 5
```
jmfranck, line 33: this should not change each time, and so should be moved into another section

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+# note that acq_time_ms is always milliseconds
+p90_us = 4.464
+deadtime_us = 10.0
+repetition_us = 12e6
+FIR_rd = 7e6
+SW_kHz = 5 
+acq_time_ms = 500. # ms
+nPoints = int(acq_time_ms*SW_kHz+0.5)
+tau_adjust_us = 0.0
+deblank_us = 1.0
```
jmfranck, line 37: this should not change every time, an so should be moved down

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+from datetime import datetime
+from SpinCore_pp.verifyParams import verifyParams
+from SpinCore_pp.power_helper import gen_powerlist
+from pyspecdata.file_saving.hdf_save_dict_to_group import hdf_save_dict_to_group
+import h5py
+# {{{ from run_Hahn_echo_mw.py
+fl = figlist_var()
+save_file=True
+# {{{ experimental parameters
+# {{{ these need to change for each sample
```
jmfranck, line 17: note this comment -- some of my conversations below refer to the fact that this block at the top should be reserved for things that change every time.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+import time
+from Instruments import power_control
+from datetime import datetime
+from SpinCore_pp.verifyParams import verifyParams
+from SpinCore_pp.power_helper import gen_powerlist
+from pyspecdata.file_saving.hdf_save_dict_to_group import hdf_save_dict_to_group
+import h5py
+# {{{ from run_Hahn_echo_mw.py
+fl = figlist_var()
+save_file=True
```
jmfranck, line 15: move this down, below the block of things that change every time.

examples/combined_ODNP.py
jmfranck: also, this does not change every time.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+# {{{ these need to change for each sample
+output_name = '150uM_TEMPOL'
+IR_postproc = 'spincore_IR_v1'
+adcOffset = 26
+carrierFreq_MHz = 14.897196
+tx_phases = r_[0.0,90.0,180.0,270.0]
+amplitude = 1.0
+nScans = 1
+nEchoes = 1
+date = datetime.now().strftime('%y%m%d')
```
jmfranck, line 26: this code does not change every time -- move it out of the block of things that changes every time.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+# all times in microseconds
+# note that acq_time_ms is always milliseconds
+p90_us = 4.464
+deadtime_us = 10.0
```
jmfranck, line 30: doesn't change every time

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+repetition_us = 12e6
+FIR_rd = 7e6
+SW_kHz = 5 
+acq_time_ms = 500. # ms
+nPoints = int(acq_time_ms*SW_kHz+0.5)
```
jmfranck, line 35: this line of code does not change every time.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+tau_adjust_us = 0.0
+deblank_us = 1.0
+tau_us = 3500
+pad_us = 0
```
jmfranck, line 39: this doesn't change every time

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+pul_prog = 'ODNP_v3'
```
jmfranck, line 40: as above, called `postproc_type` when saved to properties

jmfranck: and doesn't change every time.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,333 @@
...
+vd_list = r_[2.1e3,2.1e4,2.1e5,4.3e5,6.4e5,8.8e5,1.1e6,1.3e6,1.5e6,1.7e6,1.9e6,2.5e6,3e6,3.3e6]
+T1_powers_dB = r_[30,32,33,34]
+T1_node_names = ['FIR_30dBm','FIR_32dBm','FIR_33dBm','FIR_34dBm']
+# }}}
+#{{{Power settings
+max_power = 3 #W
+power_steps = 14
+threedown = True
```
jmfranck, line 48: these actually could change every time, so move them up

examples/combined_ODNP.py
jmfranck: edited first comment in this thread

examples/proc_combined_ODNP.py
jmfranck: no, it should be easy to implement.  I can help once everything else is resolved.

SpinCore_pp/run_field_dep_justMw.py
guinneak198: well I thought this was kind of important and we decided to increase tau to 3500 or higher only

examples/combined_ODNP.py
guinneak198: Okay so I read the pyspecdata docs and see it isn't a acquisition parameter but it's own property so I keep this structure and later set_prop('postproc_type') = IR_postproc

examples/combined_ODNP.py
guinneak198: I'm assuming you mean the SW_kHz and deadtime? the repetition delays should change for each sample

guinneak198:
Alright I reviewed and edited now back to you!

SpinCore_pp/run_field_dep_justMw.py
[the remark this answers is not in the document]
guinneak198: oops changed run_field_dep_justMw.py inside examples rather than this one inside SpinCore_pp but it should be good now!

SpinCore_pp/run_field_dep_justMw.py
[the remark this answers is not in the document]
guinneak198: should be good now

examples/proc_combined_ODNP.py
guinneak198: `[('time', '<f8'), ('Rx', '<f8'), ('power', '<f8'), ('cmd', '<i8')]` is the dtype for read_array

examples/proc_combined_ODNP.py
guinneak198: when I print the IR.getaxis('indirect') I get ```
[1.64149057e+09 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00 0.00000000e+00 0.00000000e+00 0.00000000e+00
 0.00000000e+00]
``` so yes it is a singleton dimension? should I squeeze or select the first index? what is your suggestion?

**Review by guinneak198 — COMMENTED** [review]
Reviewed JF comments and update spincore_pp/run_field_dep_justMW.py

File: SpinCore_pp/run_field_dep_justMw.py [hunk 61]
```diff
@@ -88,162 +76,126 @@ def verifyParams():
...
+                    ('pulse_TTL',p90,0.0),
+                    ('delay',tau),
+                    ('delay_TTL',deblank),
+                    ('pulse_TTL',2.0*p90,0.0),
+                    ('delay',deadtime),
+                    ('acquire',acq_time),
+                    ('delay',repetition),
+                    ('jumpto','start')
+                    ])
+                #}}}
```
jmfranck, line 141: do you ever do this? if not, delete the option to run w/out phase cycling.

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -88,162 +76,126 @@ def verifyParams():
...
+            print("\nTRANSMITTER CONFIGURED.")
+            print("***")
+            print("CONFIGURING RECEIVER...")
+            acq_time = SpinCore_pp.configureRX(SW_kHz, nPoints, nScans, nEchoes, nPhaseSteps) #ms
+            print("\nRECEIVER CONFIGURED.")
+            print("***")
+            print("\nINITIALIZING PROG BOARD...\n")
+            SpinCore_pp.init_ppg();
+            print("\nLOADING PULSE PROG...\n")
+            if phase_cycling:
```
jmfranck, line 112: maybe delete this if -- see next question

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -88,162 +76,126 @@ def verifyParams():
...
+            SpinCore_pp.configureTX(adcOffset, carrierFreq_MHz, tx_phases, amplitude, nPoints)
+            print("\nTRANSMITTER CONFIGURED.")
+            print("***")
+            print("CONFIGURING RECEIVER...")
+            acq_time = SpinCore_pp.configureRX(SW_kHz, nPoints, nScans, nEchoes, nPhaseSteps) #ms
+            print("\nRECEIVER CONFIGURED.")
+            print("***")
+            print("\nINITIALIZING PROG BOARD...\n")
+            SpinCore_pp.init_ppg();
+            print("\nLOADING PULSE PROG...\n")
```
jmfranck, line 111: convert these to debug statements

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -88,162 +76,126 @@ def verifyParams():
...
-                        data.setaxis('power',r_[powers])
-                        data.name(node_name)
-                        data.set_prop('acq_params',acq_params)
-                    data['nScans',x]['Field',B0_index]['power',0] = data_array
-                last_power = this_power
-            SpinCore_pp.stopBoard();
+total_pts = nPoints*nPhaseSteps
+assert total_pts < 2**14, "You are trying to acquire %d points (too many points) -- either change SW or acq time so nPoints x nPhaseSteps is less than 16384"%total_pts
+#{{{ppg
+def run_sweep(fieldaxis, nScans=1, B0_index, data = None):
```
jmfranck, line 91: why is this called run sweep?  This is not a sweep

jmfranck: (this might be embedded in a loop that is a sweep, but this function seems to be just a 1-off pulse sequence.

SpinCore_pp/run_field_dep_justMw.py
```diff
@@ -1,6 +1,3 @@
-from pylab import *
-from pyspecdata import *
-import os
```
jmfranck, line 3: why isn't this file inside examples?

jmfranck: in fact, there is one in examples with the same name!!  I'm going to stop here, and let you resolve that issue.

examples/combined_ODNP.py
```diff
@@ -26,28 +26,29 @@
...
 vd_list = r_[2.1e3,2.1e4,3.3e4,6.1e4,1.2e5,1.8e5,2.4e5,3e5]
-T1_powers_dB = r_[30,32,33,34]
 T1_node_names = ['FIR_%ddBm'%j for j in T1_powers_dB]
 max_power = 4 #W
 power_steps = 14
 threedown = True
 # }}}
 #{{{Power settings
 dB_settings = gen_powerlist(max_power,power_steps+1, three_down=threedown)
+T1_powers_dB = gen_powerlist(max_power, 5, three_down=False)
```
jmfranck, line 36: what I suggest here should be better / more consistent

examples/combined_ODNP.py
```diff
@@ -26,28 +26,29 @@
...
+print("T1_powers_dB",T1_powers_dB)
+print("correspond to powers in Watts",10**(T1_powers_dB/10.-3))
 myinput = input("Look ok?")
-#}}}
-time_list = [time.time()]
 if myinput.lower().startswith('n'):
     raise ValueError("you said no!!!")
+#}}}
 powers = 1e-3*10**(dB_settings/10.)
-time_list.append(time.time())
```
jmfranck: these are not used.

examples/combined_ODNP.py
```diff
@@ -56,8 +57,6 @@
 amplitude = 1.0
 deadtime_us = 10.0
 date = datetime.now().strftime('%y%m%d')
-IR_postproc = 'spincore_IR_v1'
-Ep_postproc = 'ODNP_v3'
```
jmfranck: redudant

examples/combined_ODNP.py
```diff
@@ -75,7 +74,6 @@ def run_scans(nScans, power_idx, DNP_data=None):
     ph1_cyc = r_[0,1,2,3]
     ph2_cyc = r_[0]
     nPhaseSteps = len(ph1_cyc)*len(ph2_cyc)
-    total_pts = nPoints*nPhaseSteps
```
jmfranck: redundant

examples/combined_ODNP.py
```diff
@@ -116,7 +114,6 @@ def run_scans(nScans, power_idx, DNP_data=None):
         run_scans_time_list.append(time.time())
         run_scans_names.append('get data')
         raw_data = SpinCore_pp.getData(data_length, nPoints, nEchoes, nPhaseSteps, output_name)
-        raw_data.astype(float)
```
jmfranck: doesn't do anything

examples/combined_ODNP.py
```diff
@@ -133,28 +130,28 @@ def run_scans(nScans, power_idx, DNP_data=None):
...
             data_array = nddata(array(data_array),'t')
             data_array.setaxis('t',time_axis)
-            DNP_data['indirect',power_idx]['nScans',x] = data_array
-            run_scans_time_list.append(time.time())
-            this_array = array(run_scans_time_list)
-        else:
-            time_axis = r_[0:dataPoints]/(SW_kHz*1e3)
-            data_array = nddata(array(data_array),'t')
-            data_array.setaxis('t',time_axis).set_units('t','s')
-            data_array.name('enhancement')
```
jmfranck: these 4 lines are not necessary -- beyond that, see my code comment.

SpinCore_pp/run_field_dep_justMw.py
guinneak198: took out all things that have options for phase_cycling and set the phase_Cycling as default

SpinCore_pp/run_field_dep_justMw.py
guinneak198: changed print statements inside ppg to debug statements

SpinCore_pp/run_field_dep_justMw.py
guinneak198: oops! changed name to run_scans

SpinCore_pp/run_field_dep_justMw.py
guinneak198: okay I will remove from spincore_pp and make sure the changes are carried over to the examples version!

examples/combined_ODNP.py
guinneak198: I will try to run this however, in previous runs when we were first trying this out, without these the nodenames/data would not save properly (I know that makes zero sense and it may well have been due to something else) but agreed they are unecessary

jmfranck: Don't know what you're talking about w/ the nodenames here.  This is just a list of times that is not used.

examples/proc_combined_ODNP.py
jmfranck: OK, it's a structured array, then.  So, if you wanted an array of the times, you would just do `timearray = readarray['time']`, for example.  But you don't use this, anways, so I'm deleting it.

examples/run_nutation.py
```diff
@@ -61,48 +61,30 @@ def API_sender(value):
...
-carrierFreq_MHz = 14.896101
+output_name = 'TEMPOL_capillary_probe_nutation_1'
+ph1_cyc = r_[0,2]
+ph2_cyc = r_[0,2]
+nPhaseSteps = len(ph1_cyc)*len(ph2_cyc)
+adcOffset = 39
+carrierFreq_MHz = 14.896010
 tx_phases = r_[0.0,90.0,180.0,270.0]
 amplitude = 1.0
 nScans = 1
```
jmfranck, line 72: again, go for a nice (and labeled) grouping of things that change and things that don't

examples/combined_ODNP.py
guinneak198: Okay I ran it yesterday without the time list stuff and it works!

examples/run_nutation.py
```diff
@@ -60,49 +60,34 @@ def API_sender(value):
...
 # NOTE: Number of segments is nEchoes * nPhaseSteps
 deadtime = 10.0
-repetition = 12e6
 SW_kHz = 24.0
 nPoints = 1024
 acq_time = nPoints/SW_kHz # ms
 tau_adjust = 0.0
-tau = 1000
-print("ACQUISITION TIME:",acq_time,"ms")
+tau = deadtime_us + acq_time_ms*1e3*0.5 + tau_adjust_us
```
guinneak198, line 86: do we want this hard set to 3500 like in other ppg?

guinneak198 (commented):
do we want the figures generated to be subplots so for IR processing it would be one figure with subplots for raw, phased, fitted etc?

guinneak198 (commented):
rearranged so Enhancement is run first

examples/combined_ODNP.py
```diff
@@ -0,0 +1,340 @@
...
+def run_scans_IR(vd_list, node_name, power_idx, nScans = 1, rd = FIR_rd, power_on = False, vd_data=None):
+    # (JF delete on reading): convert this to a real docstring!!
+    # nScans is number of scans you want
+    # rd is the repetition delay
+    # power_setting is what you want to run power at
+    # vd list is a list of the vd's you want to use
+    # node_name is the name of the node must specify power
+    ph1_cyc = r_[0,2]
+    ph2_cyc = r_[0,2]
+    ph3_cyc = r_[0]
```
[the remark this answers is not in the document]
jmfranck: in that case, just set this to a single number, rather than an array of length zero

examples/combined_ODNP.py
```diff
@@ -0,0 +1,343 @@
...
+print("T1_powers_dB",T1_powers_dB)
+print("correspond to powers in Watts",10**(T1_powers_dB/10.-3))
+myinput = input("Look ok?")
+if myinput.lower().startswith('n'):
+    raise ValueError("you said no!!!")
+#}}}
+powers = 1e-3*10**(dB_settings/10.)
+fl = figlist_var()
+save_file=True
+SW_kHz = 3.9
```
jmfranck, line 48: please check with @aabeaton and add a comment about what the minimum SW w/out loss is.

examples/combined_ODNP.py
```diff
@@ -0,0 +1,343 @@
...
+    raise ValueError("you said no!!!")
+#}}}
+powers = 1e-3*10**(dB_settings/10.)
+fl = figlist_var()
+save_file=True
+SW_kHz = 3.9 
+acq_time_ms = 1024. # ms
+nPoints = int(acq_time_ms*SW_kHz+0.5)
+acq_time_ms = nPoints/SW_kHz
+tau_adjust_us = 0.0
```
jmfranck, line 52: what is this?  add a comment

jmfranck: get help from @aabeaton

SpinCore_pp/ppg/echo.py
```diff
@@ -0,0 +1,78 @@
+from .. import configureRX, configureRX, init_ppg stop_ppg, runBoard, getData
+from .. import load as spincore_load
+def run_spin_echo(nScans, indirect_idx, indirect_len, DNP_data=None):
+    """run nScans and slot them into the indirect_idx index of DNP_data -- assume
+    that the first time this is run, it will be run with DNP_data=None and that
+    after that, you will pass in DNP_data
+    
+    this generates an "indirect" axis
+    """
```
jmfranck, line 9: check that imports work well

SpinCore_pp/ppg/echo.py
```diff
@@ -0,0 +1,78 @@
...
+    print("about to run run_spin_echo for",indirect_idx)
+    ph1_cyc = r_[0,1,2,3]
+    ph2_cyc = r_[0]
```
jmfranck, line 12: really, this should be a keyword argument instead, because you are setting it at the top of your example script

examples/combined_ODNP.py
```diff
@@ -0,0 +1,275 @@
...
+#{{{ enhancement curve pulse prog
+#}}}
+#{{{IR ppg
+def run_scans_IR(vd_list, node_name, power_idx, nScans = 1, rd = FIR_rd, power_on = False, vd_data=None):
+    # (JF delete on reading): convert this to a real docstring!!
+    # nScans is number of scans you want
+    # rd is the repetition delay
+    # power_setting is what you want to run power at
+    # vd list is a list of the vd's you want to use
+    # node_name is the name of the node must specify power
```
jmfranck, line 70: don't forget to move this to its own function as well

guinneak198: in ppg/IR.py

SpinCore_pp/ppg/echo.py
guinneak198: they work!

SpinCore_pp/ppg/echo.py
guinneak198: made into kwargs

SpinCore_pp/ppg/field_sweep.py
```diff
@@ -0,0 +1,67 @@
...
+                    ('phase_reset',1),
+                    ('delay_TTL',deblank_us),
+                    ('pulse_TTL',p90_us,'ph1',ph1_cyc),
+                    ('delay',tau_us),
+                    ('delay_TTL',deblank_us),
+                    ('pulse_TTL',2.0*p90_us,'ph2',ph2_cyc),
+                    ('delay',deadtime),
+                    ('acquire',acq_time),
+                    ('delay',repetition_us),
+                    ('jumpto','start')
```
jmfranck, line 40: remember to replace this with `run_spin_echo`!!

jmfranck: and test it!

examples/run_IR.py
```diff
@@ -90,24 +90,24 @@ def verifyParams():
...
-#vd_list = np.linspace(5e1,15e6,16)#5) 
-#vd_list = np.linspace(5e1,10e6,16)
-#vd_list = np.linspace(5e1,4e6,16)
-vd_list = np.linspace(5e1,6e6,16)
-#vd_list = np.linspace(5e1,3e6,15)
-#vd_list = np.linspace(5e1,8e6,16)
+#vd_list_us = np.linspace(5e1,1.8e5,32)
+#vd_list_us = np.linspace(5e1,12e6,12)
+#vd_list_us = np.linspace(5e1,15e6,16)#5) 
+#vd_list_us = np.linspace(5e1,10e6,16)
```
jmfranck, line 104: please get rid of bogus commits

examples/combined_ODNP.py
```diff
@@ -0,0 +1,206 @@
...
+        'deadtime_us', 'repetition_us', 'SW_kHz', 'nPoints', 'deblank_us',
+        'tau_us', 'MWfreq', 'acq_time_ms', 'meter_power']}
+    vd_data.set_prop('acq_params',acq_params)
+    vd_data.set_prop('postproc_type',IR_postproc)
+    vd_data.name('FIR_noPower')
+    myfilename = date+'_'+output_name+'.h5'
+    vd_data.chunk('t',['ph1','ph2','t2'],[len(IR_ph1_cyc),len(IR_ph2_cyc),-1])
+    vd_data.setaxis('ph1',IR_ph1_cyc/4)
+    vd_data.setaxis('ph2',IR_ph2_cyc/4)
+    # Need error handling (JF has posted something on this..)
```
guinneak198, line 166: what is this pertaining to?

SpinCore_pp/ppg/field_sweep.py
guinneak198: field sweep runs nicely!

examples/combined_ODNP.py
jmfranck: for writing the HDF5 file -- line 164 will refuse if the file already exists, and we don't want that to happen

jmfranck: so I had suggested that you have a try/except where it would write the file to a temporary file (say temp.h5) and ask you to move the file when it's done.  Otherwise you potentially lose a lot of data.

PR #12, opened 2022-05-09
Update ppgs

guinneak198:
use newest method of saving acquisition parameters via dictionary. 
use ppg's for IR and spin echo where applicable


examples/run_CPMG.py
```diff
@@ -12,21 +12,21 @@
 carrierFreq_MHz = 14.895548
```
guinneak198, line 12: we don't have a ppg for CPMG just yet so going forward this will need to be a new PR. I have only adjusted names of parameters to include units and the saving style of the acquisition parameters

PR #19, opened 2023-04-07
Ep spacing

guinneak198:
generate a Ep_power list that will fully cover the enhancement depending on the expected phalf.


SpinCore_pp.egg-info/PKG-INFO
```diff
@@ -0,0 +1,10 @@
+Metadata-Version: 1.0
+Name: SpinCore-pp
+Version: 0.1
+Summary: custom pulse programming language
+Home-page: http://github.com/jmfrancklab/spincore_apps
+Author: Beaton,Franck,Guinness
+Author-email: jmfranck@notgiven.com
+License: LICENSE.md
+Description: Pulse programming language for SpinCore RadioProcessor G
```
jmfranck, line 9: why is this committed???

SpinCore_pp/power_helper.py
```diff
@@ -1,8 +1,34 @@
...
+from pyspecdata import *
 "Helper functions for dealing with powers"
-
-
+#{{{ For even spacing based on phalf estimation
+def integral_as_quad(expr, lims):
+    var, a, b = lims
+    return quad(sp.lambdify(var, expr, modules='numpy'), a, b)[0]
+Emax, p, phalf, pmax = sp.symbols('E_max p p_half p_max',real=True,positive=True)
+def thisfn(sim_phalf = 0.2,sim_Emax = 30, sim_pmax = 3.16,
```
jmfranck, line 11: not a good name -- change

examples/Ep_spacing.py
```diff
@@ -0,0 +1,7 @@
+from pyspecdata import *
+import SpinCore_pp
+#config_dict = SpinCore_pp.configuration("active.ini")
+with figlist_var() as fl:
+    Ep_powers = gen_Ep_powers(sim_phalf = 0.2, sim_Emax = 10.0, sim_pmax = 3.16,
```
jmfranck, line 5: how is this supposed to work?  where is the function by this name?
also, you are not pulling from the config dict, which was the goal

jmfranck: as discussed in person, this needs to use active.ini

jmfranck: actually, changes to the config dict are going to be required, as well

jmfranck:
I know the other guy is the noapi version, and this would be the real version -- same question

guinneak198:
yes this function is already on master and used so this should be closed and the branch deleted

PR #20, opened 2023-04-10
Ep spacing noapi

guinneak198:
Ep spacing function for better power steps to fill in the initial enhancement at low powers

### DIFF
--- a/SpinCore_pp.egg-info/PKG-INFO
+++ b/SpinCore_pp.egg-info/PKG-INFO
@@ -1,0 +1,10 @@
+Metadata-Version: 1.0
+Name: SpinCore-pp
+Version: 0.1
+Summary: custom pulse programming language
+Home-page: http://github.com/jmfrancklab/spincore_apps
+Author: Beaton,Franck,Guinness
+Author-email: jmfranck@notgiven.com
+License: LICENSE.md
+Description: Pulse programming language for SpinCore RadioProcessor G
+Platform: UNKNOWN
--- a/SpinCore_pp/config_parser_fn.py
+++ b/SpinCore_pp/config_parser_fn.py
@@ -130,6 +130,12 @@ class configuration(object):
             None,
             "the range over which the dip lock will be performed",
         ),
+        "guessed_phalf": (
+            float,
+            "odnp_params",
+            None,
+            "estimated power for half saturation"
+        ),
         "adc_offset": (int, "acq_params", None, "SpinCore-specific ADC offset correction\nwe believe this is a DC offset, but are not positive"),
         "nScans": (int, "acq_params", 1, "number of scans"),
         "thermal_nScans": (
--- a/SpinCore_pp/power_helper.py
+++ b/SpinCore_pp/power_helper.py
@@ -1,8 +1,34 @@
 from numpy import *
-
+import sympy as sp
+from scipy.integrate import quad
+from pyspecdata import *
 "Helper functions for dealing with powers"
-
-
+#{{{ For even spacing based on phalf estimation
+def integral_as_quad(expr, lims):
+    var, a, b = lims
+    return quad(sp.lambdify(var, expr, modules='numpy'), a, b)[0]
+Emax, p, phalf, pmax = sp.symbols('E_max p p_half p_max',real=True,positive=True)
+def thisfn(sim_phalf = 0.2,sim_Emax = 30, sim_pmax = 3.16,
+           aspect_ratio = 1.618, p_steps = 20, fl = None):
+    p_array = r_[0:sim_pmax:200j]
+    f = 1 - Emax*(p/(p+phalf))
+    f_fn = sp.lambdify((Emax,phalf,p),f)
+    dline = sp.simplify(sp.sqrt(1+(sp.diff(f/Emax,p)*(sim_pmax/aspect_ratio))**2))
+    if fl is not None:
+        figure(figsize=(12,12/aspect_ratio))
+    length_vs_p = sp.Integral(dline,(p,0,pmax))
+    length_vs_p_fn = vectorize(
+        sp.lambdify((phalf, pmax), # no longer depends on Emax!
+                          length_vs_p,
+                          modules=[{"Integral": integral_as_quad},'sympy']),
+        excluded=[0])
+    length_data = nddata(length_vs_p_fn(sim_phalf,p_array),[-1],['p']).setaxis('p',p_array)
+    length_data.invinterp('p', linspace(0,length_data['p',-1].item(),p_steps))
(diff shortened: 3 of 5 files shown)


SpinCore_pp.egg-info/PKG-INFO
```diff
@@ -0,0 +1,10 @@
+Metadata-Version: 1.0
+Name: SpinCore-pp
+Version: 0.1
+Summary: custom pulse programming language
+Home-page: http://github.com/jmfrancklab/spincore_apps
+Author: Beaton,Franck,Guinness
+Author-email: jmfranck@notgiven.com
+License: LICENSE.md
+Description: Pulse programming language for SpinCore RadioProcessor G
+Platform: UNKNOWN
```
jmfranck, line 10: this should not be committed!  delete!

SpinCore_pp/config_parser_fn.py
```diff
@@ -130,6 +130,12 @@ class configuration(object):
             None,
             "the range over which the dip lock will be performed",
         ),
+        "guessed_phalf": (
+            float,
+            "odnp_params",
+            None,
```
jmfranck, line 136: what happens when this is set to None? does it freak out?

SpinCore_pp/power_helper.py
```diff
@@ -1,8 +1,74 @@
 from numpy import *
-
+import sympy as sp
+from scipy.integrate import quad
+from pyspecdata import *
```
jmfranck, line 4: we don't do this inside a package.  The `from numpy import *` is actually problematic, as well.  It should be:
```
from numpy import r_
import numpy as np
```

jmfranck: you can use pycharm to change the import statement and change all the uses, as well.

examples/active.ini
```diff
@@ -25,11 +25,12 @@ guessed_mhz_to_ghz = 1.5167
 
 [odnp_params]
 max_power = 3.16
-power_steps = 4
+power_steps = 14
 num_t1s = 2
 uw_dip_center_ghz = 9.820639
 uw_dip_width_ghz = 0.02
 fir_rep = 500000.0
+guessed_phalf = 0.5
```
jmfranck, line 33: I don't agree with having either of these here.  This file should have a bare minimum of parameters committed.  Any defaults should be defined above (see where I ask about `None` --> why not just give a default?)

guinneak198: but I will move phalf to sample parameters instead since it is dependent on concentration

SpinCore_pp/config_parser_fn.py
guinneak198: it will freak out if it is not in the active.ini, but after setting a default it's all good

jmfranck: why isn't this outdated?

jmfranck: as noted below, set this to a default value

SpinCore_pp/power_helper.py
jmfranck: do not `from pyspecdata import *`

jmfranck: `import pyspecdata as psp`, etc.

examples/active.ini
jmfranck: 200 μM and 9.4 mM are vastly different types of samples.  Based on that range, I think that 0.3 probably a reasonable guess.  See previous comment as to where this goes.

jmfranck: but I think it's not clear -- there should be no reference to this parameter in the committed version of active.ini, because the committed version of active.ini should be absolutely bare bones.

jmfranck: the default should be given in `SpinCore_pp/config_parser_fn.py`

SpinCore_pp/power_helper.py
jmfranck: to clarify what I'm saying -- we've learned that `from ... import *` statements in general do not play well with sphinx, so do not use them inside a package.

guinneak198: whoops didn't see that good catch

SpinCore_pp/power_helper.py
```diff
@@ -95,7 +95,7 @@ def Ep_spacing_from_phalf(
             for frac in [0.75, 0.5, 0.25]
         ]
         dB_settings = np.append(rdB_settings, append_dB)
-    return dB_settings
+    return dB_settings.real
```
jmfranck, line 98: I didn't like that it was returning complex numbers -- could generate problems later

jmfranck, line 98: (resolve when read) I didn't like that this generated complex numbers -- worried it would create problems later

examples/Ep_spacing.py
```diff
@@ -1,14 +1,17 @@
...
-    plt.xlabel('Power / W')
+    Ep_powers = Ep_spacing_from_phalf(
+        est_phalf=config_dict["guessed_phalf"] / 4,
+        max_power=config_dict["max_power"],
+        p_steps=config_dict["power_steps"],
+        fl=fl,
+    )
+    plt.title("Evenly spaced Ep curve")
+    plt.ylabel("Signal")
+    plt.xlabel("Power / W")
```
jmfranck, line 16: I just ran black here.

**But** can you please add the min dBm steps parameter as a parameter that lives inside the ODNP block (since it depends on the overall setup).  It should be included in the default active.ini, I think.

jmfranck (changes requested):
mostly just read, but note addition of new parameter requested.

examples/Ep_spacing.py
jmfranck: (it can be 1.0 vs. 0.5 vs. 0.1 depending on the source being used)

SpinCore_pp/power_helper.py
guinneak198: good catch

SpinCore_pp/power_helper.py
```diff
@@ -1,8 +1,105 @@
...
+        min_dBm_step:   float
+                        minimum stepsize that is allowed, in dBm
+        three_down:     boolean
+                        Do you want three evenly spaced powers on the way
+                        back down?
+    Return
+        ======
+        List of evenly spaced powers (dBm) for generated a nicely spaced Ep curve for a specific sample based on the estimated p_half.
+    """
+    p_array = r_[0:max_power:200j]
```
jmfranck, line 49: I noticed this guy -- is this not merged into master?  Does that mean you are not using the good spacing in your experiments?

PR #21, opened 2023-04-11
Looped thermals

guinneak198:
loop over thermals for Combined ODNP


examples/combined_ODNP.py
```diff
@@ -18,6 +20,7 @@
 
 logger = init_logging(level="debug")
 fl = figlist_var()
+thermal_scans = 4
```
jmfranck, line 23: Should this be in the configuration dictionary?

examples/combined_ODNP.py
```diff
@@ -28,6 +31,7 @@
 config_dict["date"] = date
 config_dict["odnp_counter"] += 1
 filename = f"{config_dict['date']}_{config_dict['chemical']}_{config_dict['type']}_{config_dict['odnp_counter']}"
+filename_out = filename + ".h5"
```
jmfranck, line 34: I don't understand. What is the point of breaking this across two lines of introducing a new variable?

examples/combined_ODNP.py
```diff
@@ -186,9 +190,10 @@
     )
 vd_data.rename("indirect", "vd")
 vd_data.setaxis("vd", vd_list_us * 1e-6).set_units("vd", "s")
-vd_data.chunk("t", ["ph2", "ph1", "t2"], [len(IR_ph1_cyc), len(IR_ph2_cyc), -1])
-vd_data.setaxis("ph1", IR_ph1_cyc / 4)
-vd_data.setaxis("ph2", IR_ph2_cyc / 4)
+if phase_cycling:
+    vd_data.chunk("t", ["ph2", "ph1", "t2"], [len(IR_ph1_cyc), len(IR_ph2_cyc), -1])
+    vd_data.setaxis("ph1", IR_ph1_cyc / 4)
+    vd_data.setaxis("ph2", IR_ph2_cyc / 4)
```
jmfranck, line 196: Note to self to check the definition of these variables above

examples/combined_ODNP.py
```diff
@@ -236,32 +241,48 @@
...
+                p90_us=parser_dict['p90_us'],
+                repetition_us=parser_dict['repetition_us'],
+                tau_us=parser_dict['tau_us'],
+                SW_kHz=parser_dict['SW_kHz'],
+                indirect_fields=("start_times", "stop_times"),
+                ret_data=DNP_data,
+            )
+        DNP_thermal_done = time.time()
+        time_axis_coords[j]["start_times"] = DNP_ini_time
+        time_axis_coords[j]["stop_times"] = DNP_thermal_done
```
jmfranck, line 285: Note to self to compare this to the code used elsewhere in the script. 

Also a question for Alex here.... What is the rationale where some of these are pulled directly from the dictionary and others of them are shuffled into variables first

examples/combined_ODNP.py
```diff
@@ -282,11 +303,11 @@
             raise ValueError("After 10 tries, the power has still not settled")
         time.sleep(5)
         power_settings_dBm[j] = p.get_power_setting()
-        time_axis_coords[j + 1]["start_times"] = time.time()
+        time_axis_coords[j + thermal_scans]["start_times"] = time.time()
```
jmfranck, line 306: I'm confused about why this was one before, and is essentially four now

examples/combined_ODNP.py
```diff
@@ -282,11 +303,11 @@
...
         run_spin_echo(
             nScans=config_dict["nScans"],
-            indirect_idx=j + 1,
-            indirect_len=len(powers) + 1,
+            indirect_idx=j + thermal_scans,
+            indirect_len=len(powers) + thermal_scans,
```
jmfranck, line 310: Same as previous

examples/combined_ODNP.py
```diff
@@ -298,15 +319,16 @@
...
-    DNP_data.chunk("t", ["ph1", "t2"], [len(Ep_ph1_cyc), -1])
-    DNP_data.setaxis("ph1", Ep_ph1_cyc / 4)
-    DNP_data.setaxis("nScans", r_[0 : config_dict["nScans"]])
-    DNP_data.reorder(["ph1", "nScans", "t2"])
-    DNP_data.name(config_dict["type"])
+        time_axis_coords[j + thermal_scans]["stop_times"] = time.time()
     DNP_data.set_prop("stop_time", time.time())
     DNP_data.set_prop("postproc_type", "spincore_ODNP_v4")
     DNP_data.set_prop("acq_params", config_dict.asdict())
+    DNP_data.setaxis("nScans", r_[0 : config_dict["nScans"]])
```
jmfranck, line 326: If there is only one scan, does this dimension exist? Probably it should not

examples/combined_ODNP.py
```diff
@@ -298,15 +319,16 @@
...
+    if phase_cycling:
+        DNP_data.chunk("t", ["ph1", "t2"], [len(Ep_ph1_cyc), -1])
+        DNP_data.setaxis("ph1", Ep_ph1_cyc / 4)
+        DNP_data.reorder(["ph1", "nScans", "t2"])
+    DNP_data.name(config_dict["type"])
```
jmfranck, line 331: Why does this stuff get added here? Shouldn't it have already existed once you're all done collecting data?

examples/combined_ODNP.py
```diff
@@ -320,8 +342,8 @@
             DNP_data.hdf5_write("temp_ODNP.h5", directory=target_directory)
             print(
                 "if I got this far, that probably worked -- be sure to move/rename temp_ODNP.h5 to the correct name!!"
-    print("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
-    print(("Name of saved data", echo_data.name()))
+    logger.info("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
+    logger.debug(strm("Name of saved data", echo_data.name()))
```
jmfranck, line 346: I like that you did this by the way

examples/combined_ODNP.py
```diff
@@ -362,16 +384,16 @@
             )
         vd_data.rename("indirect", "vd")
         vd_data.setaxis("vd", vd_list_us * 1e-6).set_units("vd", "s")
+        vd_data.set_prop("start_time", ini_time)
+        vd_data.set_prop("stop_time", time.time())
+        vd_data.set_prop("acq_params", config_dict.asdict())
+        vd_data.set_prop("postproc_type", "spincore_IR_v1")
+        vd_data.name(T1_node_names[j])
```
jmfranck, line 391: Same here... This got moved up. Why?

examples/combined_ODNP.py
guinneak198: we can add Ep_ph1 and IR_ph1 etc to the config file but these are pretty much never going to change so I keep them defined in this script

examples/combined_ODNP.py
guinneak198: I keep this hear as at lower concentrations we apply more nScans which makes the config file and data acquisition more robust for the user

examples/combined_ODNP.py
guinneak198: no particular reason will move back

examples/combined_ODNP.py
jmfranck: why is it getting added to this pull request?  why did it not already exist?

examples/combined_ODNP.py
```diff
@@ -31,7 +31,7 @@
 config_dict["type"] = "ODNP"
 config_dict["date"] = date
 config_dict["odnp_counter"] += 1
-filename = f"{config_dict['date']}_{config_dict['chemical']}_{config_dict['type']}_{config_dict['odnp_counter']}" + ".h5"
+filename = f"{config_dict['date']}_{config_dict['chemical']}_{config_dict['type']}_{config_dict['odnp_counter']}.h5"
```
jmfranck: (resolve on read) easier

examples/combined_ODNP.py
```diff
@@ -91,10 +91,10 @@
 )
 # }}}
 # {{{ check for file
-if os.path.exists(filename_out):
+if os.path.exists(filename):
     raise ValueError(
         "the file %s already exists, so I'm not going to let you proceed!"
-        % filename_out
+        % filename
```
jmfranck, line 99: (resolve on read) since no need for two variables, I chose the simpler name

examples/combined_ODNP.py
```diff
@@ -37,16 +37,12 @@
...
-    Ep_nPhaseSteps = 4
     IR_ph1_cyc = r_[0, 2]
     IR_ph2_cyc = r_[0, 2]
-    IR_nPhaseSteps = 4
 if not phase_cycling:
     Ep_ph1_cyc = 0.0
-    Ep_nPhaseSteps = 1
     IR_ph1_cyc = 0.0
     IR_ph2_cyc = 0.0
-    IR_nPhaseSteps = 2
```
jmfranck, line 45: not used

examples/combined_ODNP.py
```diff
@@ -101,10 +97,11 @@
 )
 # }}}
 # {{{Collect Thermals - serves as a control to compare the thermal of Ep to ensure no microwaves were leaking
+# call A to run spin echo
 control_thermal = run_spin_echo(
     nScans=config_dict["thermal_nScans"],
     indirect_idx=0,
-    indirect_len=len(powers) + 1,
+    indirect_len=len(powers) + 1,# it seems like you are saving control_thermal to a separate node vs. the main DNP data -- if that is the case, why are you telling it that it's len(powers)+1 instead of just 1 here?
```
jmfranck, line 110: needs to be addressed!

examples/combined_ODNP.py
```diff
@@ -222,11 +219,12 @@
...
         DNP_ini_time = time.time()
         if j == 0:
+            # call B to run spin echo
             DNP_data = run_spin_echo(
                 nScans=config_dict["nScans"],
                 indirect_idx=j,
                 indirect_len=len(powers) + thermal_scans,
-                ph1_cyc=Ep_ph1_cyc,
+                ph1_cyc=Ep_ph1_cyc, # this is not present in call A!
                 adcOffset=config_dict["adc_offset"],
```
jmfranck: this was repeated, ended up deleting

examples/combined_ODNP.py
```diff
@@ -156,20 +156,21 @@
...
         vd=vd,
         nScans=config_dict["thermal_nScans"],
         adcOffset=config_dict["adc_offset"],
         carrierFreq_MHz=config_dict["carrierFreq_MHz"],
         p90_us=config_dict["p90_us"],
         tau_us=config_dict["tau_us"],
         repetition_us=FIR_rep,
-        ph1_cyc=IR_ph1_cyc,
-        ph2_cyc=IR_ph2_cyc,
         SW_kHz=config_dict["SW_kHz"],
```
jmfranck, line 182: when I rearrange like this, its so the calls to these functions can be easily diff'd

examples/combined_ODNP.py
```diff
@@ -218,45 +219,27 @@
...
+            nPoints=nPoints,
+            nEchoes=config_dict["nEchoes"],
+            ph1_cyc=Ep_ph1_cyc,
+            p90_us=config_dict["p90_us"],
+            repetition_us=config_dict["repetition_us"],
+            tau_us=config_dict["tau_us"],
+            SW_kHz=config_dict["SW_kHz"],
+            indirect_fields=("start_times", "stop_times"),
+            ret_data=ret_data,
+        )
```
jmfranck, line 246: code here was vastly simplified -- does the same thing

examples/combined_ODNP.py
```diff
@@ -282,19 +265,21 @@
         power_settings_dBm[j] = p.get_power_setting()
         time_axis_coords[j + thermal_scans]["start_times"] = time.time()
         # call D to run spin echo
+        # so now you have run your thermal scans, and you are walking through the different powers?
```
jmfranck, line 273: clarify by editing comment

examples/combined_ODNP.py
```diff
@@ -74,6 +76,10 @@
     raise ValueError("you said no!!!")
 powers = 1e-3 * 10 ** (dB_settings / 10.0)
 # }}}
+# {{{ these change if we change the way the data is saved
+IR_postproc = "spincore_IR_v3" # note that you have changed the way the data is saved, and so this should change likewise!!!!
```
jmfranck, line 80: note this change

examples/combined_ODNP.py
```diff
@@ -133,19 +139,21 @@
...
-    print(
+    final_log.append(
         f"I had problems writing to the correct file {filename}, so I'm going to try to save your file to temp_ctrl.h5 in the current directory"
     )
     if os.path.exists("temp_ctrl.h5"):
-        print("There is already a temp_ctrl.h5 -- I'm removing it")
+        final_log.append("There is already a temp_ctrl.h5 -- I'm removing it")
         os.remove("temp_ctrl.h5")
-        DNP_data.hdf5_write("temp_ctrl.h5", directory=target_directory)
+        target_directory = os.path.getcwd()
```
jmfranck: if you're saving to a fallback, it's probably because you have a network problem or the directory is bad, so it should be to a local file.

examples/combined_ODNP.py
```diff
@@ -345,18 +354,17 @@
...
-        # why is this block of code moved below?
         if phase_cycling:
             vd_data.chunk("t", ["ph2", "ph1", "t2"], [len(IR_ph2_cyc), len(IR_ph1_cyc), -1])
             vd_data.setaxis("ph1", IR_ph1_cyc / 4)
             vd_data.setaxis("ph2", IR_ph2_cyc / 4)
         vd_data.setaxis("nScans", r_[0 : config_dict["nScans"]])
-        vd_data.set_prop("start_time", ini_time)
-        vd_data.set_prop("stop_time", time.time())
-        vd_data.set_prop("acq_params", config_dict.asdict())
-        vd_data.set_prop("postproc_type", "spincore_IR_v1")
```
jmfranck: I moved to better align with master

examples/combined_ODNP.py
```diff
@@ -382,3 +390,4 @@
 with h5py.File(os.path.join(target_directory, filename), "a") as f:
     log_grp = f.create_group("log")
     hdf_save_dict_to_group(log_grp, this_log.__getstate__())
+print('*'*30+'\n'+'\n'.join(final_log))
```
jmfranck, line 393: I don't like this input thing -- rather save the errors for the end.  It's faster.

examples/combined_ODNP.py
```diff
@@ -236,39 +224,37 @@
...
+        time_axis_coords[j]["start_times"] = DNP_ini_time
+        time_axis_coords[j]["stop_times"] = DNP_thermal_done
     power_settings_dBm = np.zeros_like(dB_settings)
     time.strftime("%Y-%m-%d %H:%M:%S", time.localtime(time.time()))
     for j, this_dB in enumerate(dB_settings):
         logger.debug(
             "SETTING THIS POWER", this_dB, "(", dB_settings[j - 1], powers[j], "W)"
         )
         if j == 0:
+            # why do we dip lock at the lowest power -- I don't quite understand this
```
jmfranck, line 257: why do we dip lock at the lowest power -- I don't quite understand this

I think I'm maybe just emphasizing that there is no mechanism for changing the frequency as you go up in power, but if that's the case, we already knew that, which is why we would like to implement some type of frequency lock in the server.

examples/combined_ODNP.py
```diff
@@ -371,13 +372,12 @@
         with h5py.File(
             os.path.normpath(os.path.join(target_directory, filename))
             ) as fp:
-            if nodename in fp.keys():
+            while nodename in fp.keys():
```
jmfranck, line 375: it keeps going until it finds `temp_##` that doesn't already exist

examples/combined_ODNP.py
```diff
@@ -371,13 +372,12 @@
...
                 print("this nodename already exists, so I will call it temp_%d"%j)
-                vd_data.name("temp_%d"%j)
                 nodename = "temp_%d"%j
-                vd_data.hdf5_write(f"{filename}",directory = target_directory)
-            else:
-                vd_data.hdf5_write(f"{filename}", directory=target_directory)
+                vd_data.name(nodename)
+        # hdf5_write should be outside the h5py.File with block, since it opens the file itself
+        vd_data.hdf5_write(filename, directory=target_directory)
```
jmfranck, line 380: note this.

examples/combined_ODNP.py
guinneak198: okay makes sense!

examples/combined_ODNP.py
guinneak198: oof yeah this is archaic... will update

examples/combined_ODNP.py
```diff
@@ -167,61 +164,53 @@
 ini_time = time.time()
 vd_data = None
 for vd_idx, vd in enumerate(vd_list_us):
+    # call A to run_IR
```
guinneak198, line 167: I saw this up above too... why?

examples/combined_ODNP.py
guinneak198: I went to double check the processing with load_data and v1 is the one that accepts prechunked data (like here) but still FTs with a shift

examples/combined_ODNP.py
guinneak198: It just needs to be called on the first initiation of the power control server

examples/combined_ODNP.py
guinneak198: instead of opening this file up and editing the script you simply adjust the number of scans in your active ini. One file is edited instead of touching the main ODNP script

examples/combined_ODNP.py
guinneak198: I like this

examples/combined_ODNP.py
guinneak198: sounds good

guinneak198: what are the astrix though?

examples/combined_ODNP.py
```diff
@@ -77,7 +79,7 @@
 powers = 1e-3 * 10 ** (dB_settings / 10.0)
 # }}}
 # {{{ these change if we change the way the data is saved
-IR_postproc = "spincore_IR_v3" # note that you have changed the way the data is saved, and so this should change likewise!!!!
+IR_postproc = "spincore_IR_v1" # note that you have changed the way the data is saved, and so this should change likewise!!!!
```
jmfranck, line 82: why are you rolling this back? see comment

examples/combined_ODNP.py
```diff
@@ -254,12 +251,6 @@
         logger.debug(
             "SETTING THIS POWER", this_dB, "(", dB_settings[j - 1], powers[j], "W)"
         )
-        if j == 0:
-            # why do we dip lock at the lowest power -- I don't quite understand this
-            retval = p.dip_lock(
-                config_dict['uw_dip_center_GHz'] - config_dict['uw_dip_width_GHz'] / 2,
-                config_dict['uw_dip_center_GHz'] + config_dict['uw_dip_width_GHz'] / 2,
-            )
```
jmfranck: why are you removing this?

examples/combined_ODNP.py
```diff
@@ -318,7 +309,6 @@
     # }}}
     # {{{run IR
     for j, this_dB in enumerate(T1_powers_dB):
-        # why is dip locking eliminated from here
```
jmfranck: as with above, don't delete my comment -- explain

jmfranck (changes requested):

examples/combined_ODNP.py
jmfranck: I think that you're missing that any processing script can detect *whether or not an nScans dimension exists* very very easily.

examples/combined_ODNP.py
jmfranck: does this increase the version number for the DNP data format?

examples/combined_ODNP.py
jmfranck: the change here seems to allow for the possibility that it's not phase cycled.  Does the postproc do that?

examples/combined_ODNP.py
jmfranck: OK -- we might actually want to change that -- that is a separate issue, though

examples/combined_ODNP.py
jmfranck: I label the calls to IR so that I can refer to them in the comments

jmfranck: and so when I copy/paste them to compare, I'm sure of what I'm looking at

examples/combined_ODNP.py
jmfranck: you reply to this in a different comment, and I note that the change seems to be being OK with no phase cycle -- is the postproc OK with that

examples/combined_ODNP.py
jmfranck: elsewhere you say that you need to diplock at least once, and if that's the case, don't you need this?

examples/combined_ODNP.py
guinneak198: as of right now there is no proc_script for an IR that is not phase cycled. There is one that assists a 2 x 2 phase cycle (v1) and one for a 4 x 4 (v2)

guinneak198: so yes we can make it v3

examples/combined_ODNP.py
guinneak198: I see what you're saying now

guinneak198: so I leave the setaxis for nScans when it's using thermal nscans but otherwise take out the line

examples/combined_ODNP.py
guinneak198: not sure what you mean by version number but the postproc does not change and is suited for this

examples/combined_ODNP.py
guinneak198: oh shoot yes I know AB used to do that a lot!

examples/combined_ODNP.py
guinneak198: no we need a PR in general to update postprocessing (at least the load_data.py/initial processing) but if we keep this as v3 (which doesn't exist) this can serve as a reminder?

examples/combined_ODNP.py
guinneak198: the version number is recorded in the filename and increases there - not the nodename if that's what you mean? e.i. "230413_27mM_ODNP_1/ODNP" for the first run then "230413_27mM_ODNP_2/ODNP" for the 2nd

PR #24, opened 2023-09-19
Ag cpmg

guinneak198:


examples/run_CPMG.py
```diff
@@ -31,31 +30,27 @@
 # {{{set phase cycling
 phase_cycling = True
 if phase_cycling:
-    ph1_cyc = r_[0, 1, 2, 3]
-    nPhaseSteps = 4
+    ph1_cyc = r_[0, 2, 1, 3, 2, 4, 3, 5]
```
jmfranck, line 33: we want this:
```
In [59]: for k in range(4):
    ...:     for j in range(2):
    ...:         print("90 pulse phase is",(j*2+k)%4,"180 pulse phase is",(k+1)%4) # modulus, converts 4 to 0
    ...: 
90 pulse phase is 0 180 pulse phase is 1
90 pulse phase is 2 180 pulse phase is 1
90 pulse phase is 1 180 pulse phase is 2
90 pulse phase is 3 180 pulse phase is 2
90 pulse phase is 2 180 pulse phase is 3
90 pulse phase is 0 180 pulse phase is 3
90 pulse phase is 3 180 pulse phase is 0
90 pulse phase is 1 180 pulse phase is 0
```

jmfranck: this can be achieved with list comprehensions like so:
```
ph1 = np.array([(j*2+k)%4  for k in range (4) for j in range(2)]) # inner loop comes last
ph2 = np.array([(k+1)%4  for k in range (4) for j in range(2)]) # inner loop comes last
```

examples/run_CPMG.py
```diff
@@ -66,56 +61,54 @@
...
-    ret_data = None,
-)
+data = generic(
+        ppg_list = [
+                ("phase_reset", 1),
+                ("delay_TTL", config_dict['deblank_us']),
+                ("pulse_TTL", config_dict['p90_us'], "ph1", ph1_cyc),
+                ("delay", tau_us),
+                ("delay_TTL", config_dict['deblank_us']),
+                ("pulse_TTL", 2.0 * config_dict['p90_us'], "ph2", ph2_cyc),
```
jmfranck, line 71: because we are doing our cycling manually, above, you want to *also* call this "ph1" (see the docstring for `apply_cycles`)

jmfranck: actually,  to avoid confusion, I would just call this `ph_cycle`.'

and, since this is going into examples in the docstring on the top, you want to reference :func:`apply_cycles`

examples/run_CPMG.py
```diff
@@ -4,7 +4,7 @@
 
 This script will perform a standard CPMG experiment. 
 In order to form a symmetric echo, a padding time is added before 
-and after your tau through a series of delays. 
+and after your tau.
```
jmfranck, line 7: no, link to figure -- use rst to do so.

jmfranck, line 7: no, link to figure -- use rst to do so.

examples/run_CPMG.py
```diff
@@ -31,31 +30,25 @@
 # {{{set phase cycling
 phase_cycling = True
 if phase_cycling:
-    ph1_cyc = r_[0, 1, 2, 3]
-    nPhaseSteps = 4
+    ph1_cyc = array([(j*2+k)%4 for k in range(4) for j in range(2)])
+    ph2_cyc = array([(k+1)%4 for k in range(4) for j in range(2)])
+    nPhaseSteps = 8
```
jmfranck, line 35: no: (1) are you sure you need phase steps and (2) why isn't it len of previous???

examples/run_CPMG.py
```diff
@@ -31,31 +30,25 @@
...
-assert(pad_start > 0), "tau_extra_us must be set to more than deadtime and more than deblank!"
-assert (pad_end > 0), "tau_extra_us must be set to more than deadtime and more than deblank!"
-twice_tau_echo_us = (  # the period between 180 pulses
-    config_dict["tau_extra_us"] * 2 + config_dict["acq_time_ms"] * 1e3
+# {{{symmetric tau
+marker = 1.0
+jumpto = 1.0
+pad_end = config_dict['deadtime_us'] - config_dict['deblank_us'] - marker - jumpto
+twice_tau_echo_us = (  # the period between end of first 180 pulse and start of next
+    config_dict['deadtime_us'] + config_dict["acq_time_ms"] * 1e3 + pad_end + marker + config_dict['deblank_us']
```
jmfranck, line 46: - are you sure that you are rounding acq_time_ms to an integral number of dwell times? (I have written code to do this, so it should be available somewhere).  That needs to be done in this pulse sequence.  If so, you should add a comment where that is done, referencing that it's necessary to make this work correctly.
- base on previous, why not just multiply deadtime by 2? (Since you are forcing the stuff on the right to be equal to the deadtime.

examples/run_CPMG.py
```diff
@@ -31,31 +30,25 @@
...
 )
-# now twice_tau_echo_us/2.0 is τ_echo, so I need to subtract the extra delays
-# imposed by the ppg to determine the remaining τ that I need
-config_dict["tau_us"] = twice_tau_echo_us / 2.0 - (
+config_dict["tau_us"] = twice_tau_echo_us / 2.0 + (
```
jmfranck, line 48: this doesn't work.  You should define evolution during the pulse before this, and subtract it here.  Also, as noted w.r.t. the diagram, are you sure there is no delay either after the 90 or before the 180 that you need to include here?

I'm stopping here -- this PR is not ready.

examples/run_CPMG.py
guinneak198: see lines 55-58 where I verify the number of total points using nPhaseSteps - not sure what you are asking for #2

examples/run_CPMG.py
guinneak198: (1) twice_tau echo does not include the tau_evol so it needs to be added
(2) there is no deadtime involved with the tau only a deblank prior to the 180 pulse which I added in as well

examples/run_CPMG.py
guinneak198: done

examples/CPMG.rst
```diff
@@ -0,0 +1 @@
+.. image:: CPMG_ppg.jpg
```
jmfranck, line 1: don't make a separate file -- put this right in the docstring.

examples/run_CPMG.py
```diff
@@ -31,91 +31,71 @@
 # {{{set phase cycling
 phase_cycling = True
 if phase_cycling:
-    ph1_cyc = r_[0, 1, 2, 3]
-    nPhaseSteps = 4
+    Nph_overall = r_[0,1,2,3]
```
jmfranck, line 34: call these `ph_overall` and `ph_diff` (`N` doesn't make sense -- these are phase lists -- when we were talking about `N` it was when it was the length of the list -- like `nPhaseSteps`)

examples/run_CPMG.py
```diff
@@ -31,91 +31,71 @@
...
+    Nph_diff = r_[0,2]
+    ph1_cyc = array([(j+k)%4 for k in Nph_overall for j in Nph_diff])
+    ph2_cyc = array([(k+1)%4 for k in Nph_overall for j in Nph_diff])
+    nPhaseSteps = len(Nph_overall) * len(Nph_diff)
 if not phase_cycling:
     ph1_cyc = 0.0
+    ph2_cyc = 0.0
+    Nph_overall = r_[0]
+    Nph_diff = r_[0]
     nPhaseSteps = 1
```
jmfranck, line 44: none of this should be here.  Running without phase cycling is not something we want to do.

examples/run_CPMG.py
```diff
@@ -31,91 +31,71 @@
...
-config_dict["tau_us"] = twice_tau_echo_us / 2.0 - (
-    2
-    * config_dict["p90_us"]
-    / pi  # evolution during pulse -- see eq 6 of coherence paper
-    + config_dict["deadtime_us"]  # following 90
-    + config_dict["deblank_us"]  # before 180
-)
+# {{{symmetric tau
+marker = 1.0
+jumpto = 1.0
```
jmfranck, line 48: I think call this `short_delay_us`, and use for both.  This solves two problems:

- do we really need two variables, since we want them both as short as possible?
- these are not named in a way that makes it clear what their units are.

examples/run_CPMG.py
```diff
@@ -31,91 +31,71 @@
...
+tau_evol = 2*config_dict['p90_us']/pi #evolution during pulse -- see eq 6 of coherence paper
```
jmfranck, line 49: similarly, `tau_evol_us`

**btw, you do this with**
`:s/\<tau_evol\>/&_us/g` where `&` means "the whole pattern" and for anything else you don't understand, see [this](https://remram44.github.io/regex-cheatsheet/regex.html)

examples/run_CPMG.py
```diff
@@ -31,91 +31,71 @@
...
+pad_end = config_dict['deadtime_us'] - config_dict['deblank_us'] - marker - jumpto
```
jmfranck, line 50: `pad_end_us`

examples/run_CPMG.py
```diff
@@ -31,91 +31,71 @@
...
+twice_tau_echo_us = config_dict['acq_time_ms'] + (2 * config_dict['deadtime_us'])# the period between end of first 180 pulse and start of next
+config_dict["tau_us"] = twice_tau_echo_us / 2.0 - tau_evol - config_dict['deblank_us']
 # }}}
 # {{{check total points
-total_pts = nPoints * nPhaseSteps
+total_pts = nPoints * nPhaseSteps * config_dict['nEchoes']
```
jmfranck, line 55: don't forget to run black

examples/run_CPMG.py
```diff
@@ -31,91 +31,71 @@
...
-    data.rename("t","t2")
-    fl.next("Raw - time")
-    fl.image(
-        data.C.mean("nScans"))
-    data.reorder("t2", first=False)
-    for_plot = echo_data.C
-    for_plot.ft('t2',shift=True)
-    fl.next('FTed data')
-    fl.image(for_plot)
+data.chunk("t", ["ph_overall", "ph_diff", "tE", "t2"], [len(Nph_overall), len(Nph_diff), config_dict["nEchoes"], nPoints])
```
jmfranck, line 96: The generic ppg creates an `nScans` dimension, and does not squeeze.  You also write the code above allowing for nScans.  So, you should allow for that here, *as well as* when you visualize.

examples/run_CPMG.py
```diff
@@ -149,4 +129,14 @@
 print("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
 print(("Name of saved data", data.name()))
 config_dict.write()
+#{{{ Image raw data
+fl.next("Raw - time")
+fl.image(data)
```
jmfranck, line 134: mean over `nScans` dimension.

jmfranck (changes requested):

examples/run_CPMG.py
guinneak198: mentioned in docstring

examples/run_CPMG.py
guinneak198: line 24 handles the rounding acq_time

guinneak198: fixed twice tau to be 2xdeadtime +acq

examples/run_CPMG.py
guinneak198: done and discussed

examples/CPMG.rst
guinneak198: done

examples/run_CPMG.py
guinneak198: okay so to avoid making an extra line (e.g.`nph_overall = len(ph_overall)`) I am just calling` len(ph_overall) `where necessary (eg. setting the axis

examples/run_CPMG.py
guinneak198: fixed

examples/run_CPMG.py
guinneak198: "allow for that here" to me means include an nScans in the chunking and mean for imaging.... so I did that but I'm not sure if that was what you meant?

guinneak198: based on your next comment I think that's right

examples/run_CPMG.py
```diff
@@ -2,24 +2,28 @@
 CPMG
 ====
 
+..image:: CPMG_ppg.jpg
+
 This script will perform a standard CPMG experiment. 
 In order to form a symmetric echo, a padding time is added before 
-and after your tau through a series of delays. 
+and after your tau. Both the initial ninety pulse and
```
jmfranck, line 9: You only add after, right?

examples/run_CPMG.py
```diff
@@ -31,91 +35,79 @@
...
-    p90_us = config_dict["p90_us"],
-    repetition_us = config_dict["repetition_us"],
-    tau_us = config_dict["tau_us"],
-    SW_kHz = config_dict["SW_kHz"],
-    pad_start_us = pad_start,
-    pad_end_us = pad_end,
-    ret_data = None,
+data = generic(
+    ppg_list=[
+        ("marker", "start", 1),
```
jmfranck, line 65: Is this normal? Shouldn't there be a mention of nScans here?

examples/run_CPMG.py
```diff
@@ -31,91 +35,79 @@
...
-    data.reorder("t2", first=False)
-    for_plot = echo_data.C
-    for_plot.ft('t2',shift=True)
-    fl.next('FTed data')
-    fl.image(for_plot)
+data.chunk(
+    "t",
+    ["nScans", "ph_overall", "ph_diff", "tE", "t2"],
+    [
+        len(config_dict["nScans"]),
```
jmfranck, line 101: No len here

examples/run_CPMG.py
```diff
@@ -31,91 +35,79 @@
...
+        len(ph_overall),
+        len(ph_diff),
+        config_dict["nEchoes"],
+        nPoints,
+    ],
+)
+data.setaxis("nScans", r_[0 : len(config_dict["nScans"])])
+data.setaxis("ph_overall", ph_overall / len(ph_overall))
+data.setaxis("ph_diff", ph_diff / len(ph_diff))
```
jmfranck, line 110: These are both just divided by 4

examples/run_CPMG.py
```diff
@@ -149,4 +141,15 @@
...
+with figlist_var() as fl:
+    fl.next("Raw - time")
+    fl.image(data.mean("nScans"))
+    data.reorder("t2", first=False)
+    for_plot = data.C
+    for_plot.ft("t2", shift=True)
+    for_plot.ft(["ph_overall", "ph_diff"], unitary=True)
+    fl.next("FTed data")
+    fl.image(for_plot.mean("nScans"))
+    fl.show()
```
jmfranck, line 154: If you use a with block, you do not need a show at the end

jmfranck (changes requested):

examples/run_CPMG.py
guinneak198: yes good catch!

examples/run_CPMG.py
guinneak198: line 86 below

examples/run_CPMG.py
guinneak198: fixed

examples/run_CPMG.py
guinneak198: want to discuss why in gm or later but changed

examples/run_CPMG.py
guinneak198: cool wasn't sure!

examples/run_CPMG.py
jmfranck: I think `examples/CPMG_ppg.jpg`

examples/run_CPMG.py
```diff
@@ -2,24 +2,28 @@
...
 
+..image:: CPMG_ppg.jpg
+
 This script will perform a standard CPMG experiment. 
 In order to form a symmetric echo, a padding time is added before 
-and after your tau through a series of delays. 
+and after your tau. Both the initial ninety pulse and 
+subsequent 180 pulses are phase cycled together,
+with a two step phase cycle on the ninety pulse (see
+:func: apply_cycles in SpinCore_pp.py)
```
jmfranck, line 12: you're missing backticks here surrounding the apply_cycles

examples/run_CPMG.py
```diff
@@ -149,4 +141,14 @@
...
+with figlist_var() as fl:
+    fl.next("Raw - time")
+    fl.image(data.mean("nScans"))
+    data.reorder("t2", first=False)
+    for_plot = data.C
+    for_plot.ft("t2", shift=True)
+    for_plot.ft(["ph_overall", "ph_diff"], unitary=True)
+    fl.next("FTed data")
+    fl.image(for_plot.mean("nScans"))
+# }}}
```
jmfranck, line 154: we also discussed how you need to update `generic.py` as well

examples/run_CPMG.py
guinneak198: I think fixed? check syntax

SpinCore_pp/ppg/generic.py
```diff
@@ -101,14 +102,20 @@ def generic(
...
-    # }}}
+    # {{{ tuples with 4 elements are pulses, where the 4th element is the phase cycle and tuples containing marker and echo label are cycled by nEchoes
+    phcyc_lens = {}
+    nEchoes = 1
+    for thiselem in ppg_list:
+        print(len(thiselem))
+        if len(thiselem) > 3:
+            phcyc_lens[thiselem[2]] = len(thiselem[3])
+        if len(thiselem)>2 and thiselem[0] == 'marker' and thiselem[1] == 'echo_label':  
+            print(thiselem)
```
jmfranck, line 113: print statements are diagnostic -- remove them.

SpinCore_pp/ppg/generic.py
```diff
@@ -117,8 +124,10 @@ def generic(
         configureTX(adcOffset, carrierFreq_MHz, tx_phases, amplitude, nPoints)
         run_scans_time_list.append(time.time())
         run_scans_names.append("configure Rx")
-        check = configureRX(SW_kHz, nPoints, nScans, nEchoes, nPhaseSteps)
-        assert acq_time_ms == check
+        check = round(configureRX(SW_kHz, nPoints, nScans, nEchoes, int(nPhaseSteps)),1)
+        print(check)
+        print(acq_time_ms)
```
jmfranck, line 129: more print statements that are diagnostic -- remove.  If you really want to check something, then use an assert

SpinCore_pp/ppg/generic.py
```diff
@@ -101,14 +102,20 @@ def generic(
...
+        print(len(thiselem))
+        if len(thiselem) > 3:
+            phcyc_lens[thiselem[2]] = len(thiselem[3])
+        if len(thiselem)>2 and thiselem[0] == 'marker' and thiselem[1] == 'echo_label':  
+            print(thiselem)
+            nEchoes = thiselem[2]+1
+            print(nEchoes)
+    nPhaseSteps_min = 1
+    ph_lens = list(phcyc_lens.values())
+    nPhaseSteps = prod(ph_lens)
```
jmfranck, line 118: could this just be one line?

SpinCore_pp/ppg/generic.py
```diff
@@ -133,7 +142,8 @@ def generic(
         runBoard()
         run_scans_time_list.append(time.time())
         run_scans_names.append("get data")
-        raw_data = getData(data_length, nPoints, nEchoes, nPhaseSteps)
+        raw_data = getData(int(data_length), nPoints, nEchoes, int(nPhaseSteps),"thisIsntFixedYet")
```
jmfranck, line 145: this is absolutely fixed!

jmfranck (commented):
stopping -- seems like this might be old?

jmfranck:
this is merged into CPMG_tests, which is more manageable.

PR #26, opened 2023-11-06
Ag debug diplock

guinneak198:
we had a dip lock at the end to avoid the error when we go up in steps >3 dB. Here I now have a loop to step up in increments of 2 dB instead to avoid extra dip locking

### DIFF
--- a/examples/combined_ODNP.py
+++ b/examples/combined_ODNP.py
@@ -108,6 +108,52 @@ input(
     "B12 needs to be unplugged and turned off for the thermal! Don't have the power server running just yet"
 )
 # }}}
+# {{{Collect Thermals - serves as a control to compare the thermal of Ep to ensure no microwaves were leaking
+# call A to run spin echo
+control_thermal = run_spin_echo(
+    nScans=config_dict["thermal_nScans"],
+    indirect_idx=0,
+    indirect_len=1,
+    ph1_cyc=Ep_ph1_cyc,
+    adcOffset=config_dict["adc_offset"],
+    carrierFreq_MHz=config_dict["carrierFreq_MHz"],
+    nPoints=nPoints,
+    nEchoes=config_dict["nEchoes"],
+    p90_us=config_dict["p90_us"],
+    repetition_us=config_dict["repetition_us"],
+    tau_us=config_dict["tau_us"],
+    SW_kHz=config_dict["SW_kHz"],
+    ret_data=None,
+) 
+if config_dict["thermal_nScans"] > 1:
+    control_thermal.setaxis("nScans", r_[0 : config_dict["thermal_nScans"]])
+if phase_cycling:
+    control_thermal.chunk("t", ["ph1", "t2"], [len(Ep_ph1_cyc), -1])
+    control_thermal.setaxis("ph1", Ep_ph1_cyc / 4)
+    control_thermal.reorder(["ph1", "nScans", "t2"])
+control_thermal.name("control_thermal")
+control_thermal.set_prop("postproc_type", Ep_postproc)
+control_thermal.set_prop("acq_params", config_dict.asdict())
+control_thermal.name("control_thermal")
+nodename = control_thermal.name()
+# {{{ on first write, if we can't access the directory, write to a temp file
+try:
+    control_thermal.hdf5_write(filename, directory=target_directory)
+except:
+    final_log.append(
+        f"I had problems writing to the correct file {filename}, so I'm going to try to save your file to temp_ctrl.h5 in the current directory"
+    )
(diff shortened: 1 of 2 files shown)


examples/combined_ODNP.py
```diff
@@ -230,7 +230,7 @@
         DNP_data = run_spin_echo(
             nScans=config_dict["nScans"],
             indirect_idx=j,
-            indirect_len=len(powers) + thermal_scans,
+            indirect_len=len(powers) + config_dict['thermal_nScans'],
```
guinneak198, line 233: config dict parameter we skipped by earlier

examples/combined_ODNP.py
```diff
@@ -309,13 +309,27 @@
             os.remove("temp_ODNP.h5")
             DNP_data.hdf5_write(filename, directory=target_directory)
             final_log.append(
-                "if I got this far, that probably worked -- be sure to move/rename temp_ODNP.h5 to the correct name!!"
+                "if I got this far, that probably worked -- be sure to move/rename temp_ODNP.h5 to the correct name!!")
```
guinneak198, line 312: was missing a closing parenthesis

examples/combined_ODNP.py
```diff
@@ -309,13 +309,27 @@
...
     logger.info("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
-    logger.debug(strm("Name of saved data", echo_data.name()))
+    logger.debug(strm("Name of saved data", DNP_data.name()))
```
guinneak198, line 314: there is no echo_data. It's named DNP_Data not sure how this snuck by

examples/combined_ODNP.py
```diff
@@ -309,13 +309,27 @@
...
     # }}}
     # {{{run IR
+    last_dB_setting = 10
     for j, this_dB in enumerate(T1_powers_dB):
-        p.set_power(this_dB)
+        # {{{ make small steps in power if needed
+        if j == 0:
+            p.set_power(this_dB)
```
guinneak198, line 321: gen_powerlist returns a list of powers starting at 0 dB. I think it might be useful to have an extra FIR_0dBm to compare with FIR_noPower. Of course this is my opinion and if you think otherwise I can edit to not include FIR_0dBm

examples/combined_ODNP.py
```diff
@@ -378,10 +391,6 @@
         vd_data.hdf5_write(filename, directory=target_directory)
         print("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
         print(("Name of saved data", vd_data.name()))
-    final_frq = p.dip_lock(
-        config_dict["uw_dip_center_GHz"] - config_dict["uw_dip_width_GHz"] / 2,
-        config_dict["uw_dip_center_GHz"] + config_dict["uw_dip_width_GHz"] / 2,
-    )
```
guinneak198, line 384: we don't need this anymore!

examples/combined_ODNP.py
```diff
@@ -309,13 +309,26 @@
...
+            p.set_power(this_dB)
+        else:    
+            if this_dB - last_dB_setting > 3:
+                smallstep_dB = last_dB_setting + 2
+                while smallstep_dB + 2 < this_dB:
+                    p.set_power(smallstep_dB)
+                    smallstep_dB += 2
+                    last_dB_setting = this_dB
+                p.set_power(this_dB)
+        last_dB_setting = this_dB
```
guinneak198, line 330: written by JF debugged by AG - works

examples/combined_ODNP.py
```diff
@@ -309,13 +309,26 @@
...
-    logger.debug(strm("Name of saved data", echo_data.name()))
+    logger.debug(strm("Name of saved data", DNP_data.name()))
     # }}}
     # {{{run IR
+    last_dB_setting = 10
     for j, this_dB in enumerate(T1_powers_dB):
-        p.set_power(this_dB)
+        # {{{ make small steps in power if needed
+        if j == 0:
+            p.set_power(this_dB)
```
jmfranck, line 321: The part in the if block shouldn't be needed. The else should work given that you set last power to 10 above

examples/combined_ODNP.py
jmfranck, line 330: This line should be at the same level as the previous line, and the line two up should be deleted

jmfranck (changes requested):

examples/combined_ODNP.py
guinneak198: I think you were referring to the `if this_dB == 0`? so I edited and tested with ipython and this runs well

jmfranck: I'm talking about `if j==0`

PR #27, opened 2023-11-10
this was successfully run 11/09

guinneak198:
debugging errors that popped up when I tried to run

### DIFF
--- a/examples/combined_ODNP.py
+++ b/examples/combined_ODNP.py
@@ -65,7 +65,7 @@ config_dict['FIR_rep'] = FIR_rep
 dB_settings = Ep_spacing_from_phalf(
     est_phalf = config_dict['guessed_phalf'],
     max_power = config_dict["max_power"], 
-    config_dict["power_steps"] + 1, 
+    p_steps = config_dict["power_steps"] + 1, 
     min_dBm_step = config_dict['min_dBm_step'],
     three_down=True
 )
@@ -90,7 +90,7 @@ Ep_postproc = "spincore_ODNP_v3"
 total_points = len(Ep_ph1_cyc) * nPoints
 assert total_points < 2 ** 14, (
     "For Ep: You are trying to acquire %d points (too many points) -- either change SW or acq time so nPoints x nPhaseSteps is less than 16384\nyou could try reducing the acq_time_ms to %f"
-    % total_pts
+    % total_points
 )
 total_pts = len(IR_ph2_cyc) * len(IR_ph1_cyc) * nPoints
 assert total_pts < 2 ** 14, (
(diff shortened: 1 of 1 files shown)


examples/combined_ODNP.py
```diff
@@ -244,6 +244,8 @@
             ret_data=DNP_data,
         )
         DNP_thermal_done = time.time()
+        if j == 0:
+            time_axis_coords = DNP_data.getaxis("indirect")
```
guinneak198, line 248: time_axis_coords needs to be defined at some point

examples/combined_ODNP.py
```diff
@@ -311,7 +313,7 @@
             final_log.append(
                 "if I got this far, that probably worked -- be sure to move/rename temp_ODNP.h5 to the correct name!!"
     logger.info("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
-    logger.debug(strm("Name of saved data", echo_data.name()))
+    logger.debug(strm("Name of saved data", DNP_data.name()))
```
guinneak198, line 316: no such thing as echo_data

examples/combined_ODNP.py
```diff
@@ -378,14 +380,10 @@
...
         print(("Name of saved data", vd_data.name()))
-    final_frq = p.dip_lock(
-        config_dict["uw_dip_center_GHz"] - config_dict["uw_dip_width_GHz"] / 2,
-        config_dict["uw_dip_center_GHz"] + config_dict["uw_dip_width_GHz"] / 2,
-    )
     this_log = p.stop_log()
 # }}}
 config_dict.write()
-with h5py.File(os.path.join(target_directory, filename), "a") as f:
+with h5py.File(os.path.normpath(os.path.join(target_directory, filename)), "a") as f:
```
guinneak198, line 386: using this the log saved successfullly in the right h5 file

examples/combined_ODNP.py
```diff
@@ -378,14 +380,10 @@
         vd_data.hdf5_write(filename, directory=target_directory)
         print("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
         print(("Name of saved data", vd_data.name()))
-    final_frq = p.dip_lock(
-        config_dict["uw_dip_center_GHz"] - config_dict["uw_dip_width_GHz"] / 2,
-        config_dict["uw_dip_center_GHz"] + config_dict["uw_dip_width_GHz"] / 2,
-    )
```
guinneak198, line 384: not needed!

examples/combined_ODNP.py
```diff
@@ -65,7 +65,7 @@
 dB_settings = Ep_spacing_from_phalf(
     est_phalf = config_dict['guessed_phalf'],
     max_power = config_dict["max_power"], 
-    config_dict["power_steps"] + 1, 
+    p_steps = config_dict["power_steps"] + 1,
```
jmfranck, line 68: why are you feeding +1 here? Why not just config_dict['power_steps']?

jmfranck (changes requested):
one question

examples/combined_ODNP.py
guinneak198: this was pulled in from previous runs (prior to when we had thermals in a separate loop to accomodate a power axis long enough for including the thermals. - but you point out that this is no longer needed

PR #28, opened 2023-11-14
No ir half power

wfkincaid:
This is a small fix to not have powers in the generated IR list to *.5 dB steps, because the HP won't settle on values to the *.5 power.

### DIFF
--- a/examples/combined_ODNP.py
+++ b/examples/combined_ODNP.py
@@ -72,7 +72,7 @@ dB_settings = Ep_spacing_from_phalf(
 T1_powers_dB = gen_powerlist(
     config_dict["max_power"], config_dict["num_T1s"], three_down=False
 )
-T1_powers_dB = [int(a) for a in T1_powers_dB]
+T1_powers_dB = np.floor(T1_powers_dB)
 T1_node_names = ["FIR_%ddBm" % j for j in T1_powers_dB]
 logger.info("dB_settings", dB_settings)
 logger.info("correspond to powers in Watts", 10 ** (dB_settings / 10.0 - 3))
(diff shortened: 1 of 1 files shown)


examples/combined_ODNP.py
```diff
@@ -72,6 +72,7 @@
 T1_powers_dB = gen_powerlist(
     config_dict["max_power"], config_dict["num_T1s"], three_down=False
 )
+T1_powers_dB = np.floor(T1_powers_dB)
```
jmfranck, line 75: this works specifically for this purpose, but it's better to set the `min_dBm_step` argument of `gen_powerlist`.  Here, you can just click on the function name, and see the docstring.

jmfranck (changes requested):

examples/combined_ODNP.py
jmfranck: note that you can also test that your solution works by just writing a small script (not committing to the repo -- just so you can see it works) that prints the result of `gen_powerlist`

examples/combined_ODNP.py
```diff
@@ -70,7 +70,7 @@
     three_down=True
 )
 T1_powers_dB = gen_powerlist(
-    config_dict["max_power"], config_dict["num_T1s"], three_down=False
+    config_dict["max_power"], config_dict["num_T1s"], config_dict["min_dBm_step"], three_down=False
```
jmfranck, line 73: I think this is a keyword argument. While you can skip the keyword and rely on order, for clarity, it's best to give the keyword.

jmfranck (changes requested):

examples/combined_ODNP.py
wfkincaid: so it would be better to have
```gen_powerlist(
max_power = config_dict["max_power"], steps = configdict["num_T1s"], ...
```?

PR #29, opened 2024-01-31
90 pulse stuff

guinneak198:
calibrated 90 and 180 pulses

### DIFF
--- a/SpinCore_pp/ppg/echo.py
+++ b/SpinCore_pp/ppg/echo.py
@@ -86,11 +86,12 @@ def run_spin_echo(
                     returned data from previous run or `None` for the first run.
     """
     assert nEchoes == 1, "you must only choose nEchoes=1"
-    desired_p90_us = p90_us
-    prog_p90_us = prog_plen(desired_p90_us)
-    prog_p180_us = prog_plen(2*desired_p90_us)
-    print("You said you want an actual p90 of %d us and an actual p180 of %d"%(desired_p90_us,2*desired_p90_us))
-    print("So I am sending the SC a p90 time of %d and a p180 time of %d"%(prog_p90_us,prog_p180_us)
+    # take the desired p90 and p180
+    # (2*desired_p90) and convert to what needs to
+    # be programmed in order to get the desired
+    # times
+    prog_p90_us = prog_plen(p90_us)
+    prog_p180_us = prog_plen(2 * p90_us)
     tx_phases = r_[0.0, 90.0, 180.0, 270.0]
     nPhaseSteps = len(ph1_cyc) * len(ph2_cyc)
     data_length = 2 * nPoints * nEchoes * nPhaseSteps
--- a/SpinCore_pp/ppg/echo.py
+++ b/SpinCore_pp/ppg/echo.py
@@ -87,10 +87,9 @@ def run_spin_echo(
     """
     assert nEchoes == 1, "you must only choose nEchoes=1"
     desired_p90_us = p90_us
-    desired_p180_us = 2*p90_us
     prog_p90_us = prog_plen(desired_p90_us)
-    prog_p180_us = prog_plen(desired_180_us)
-    print("You said you want an actual p90 of %d us and an actual p180 of %d"%(desired_p90_us,desired_p180_us))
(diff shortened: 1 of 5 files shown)


examples/90time_conv.py
```diff
@@ -0,0 +1,19 @@
...
+from pyspecdata import *
+from SpinCore_pp.pulse_length_conv import prog_plen
+
+desired_plen = r_[0:25:25j]
+prog_plen = prog_plen(desired_plen)
+with figlist_var() as fl:
+    fl.next("Actual vs programmed plen")
+    plt.plot(prog_plen, desired_plen, "o")
+    plt.ylabel(r"actual plen / $\mu$s")
+    plt.xlabel(r"programmed plen / $\mu$s")
```
guinneak198, line 19: might need to pull in the API disabler stuff to run this... but I ran on the spincore computer and it runs

SpinCore_pp/pulse_length_conv.py
```diff
@@ -0,0 +1,61 @@
+from pylab import *
+from pyspecdata import *
```
jmfranck, line 2: No import * inside the package

SpinCore_pp/pulse_length_conv.py
```diff
@@ -0,0 +1,61 @@
...
+        (10.5, 98.6599, 245.213),
+        (11, 105.327, 258.004),
+        (11.5, 112.985, 272.776),
+        (12, 120.264, 286.884),
+        (12.5, 126.251, 299.009),
+    ]
+    # neat JF trick for organizing these data points
+    prog90, act90, act180 = map(array, zip(*datapoints))
+    # }}}
+    plen_prog = r_[0, prog90, 2 * prog90]  # array of programmed pulse lengths
```
jmfranck, line 45: I don't understand this line

jmfranck: After reading below, I think I understand it, but it could use a better comment where you talk about your array above was generated from a spin echo pulse sequence, and so you have the 90 and 180 times listed separately, where here you gather them into a single array so that you are ready to interpolate

SpinCore_pp/pulse_length_conv.py
```diff
@@ -0,0 +1,61 @@
...
+    calibration_data.sort("plen")
+    c = calibration_data.polyfit(
+        "plen", order=10
+    )  # fit the programmed vs actual lengths to a polynomial
+    # {{{ Make an array out of the fitting coefficients from the polyfit
+    coeffs = np.zeros(len(c))
+    for j in range(len(c)):
+        coeffs[j] = c[j]
+    # }}}
+    return polyval(coeffs[::-1],desired_actual)
```
jmfranck, line 61: There's a pyspecdata function to do this so that you don't have to flip the order of the coefficients

jmfranck (changes requested):

SpinCore_pp/pulse_length_conv.py
guinneak198: commented question for single p90 value here https://jmfrancklab.slack.com/archives/CKZ9UMEH2/p1706798163404309

guinneak198: discussed an we are going to keep polyval for single points

SpinCore_pp/ppg/echo.py
```diff
@@ -86,6 +86,11 @@ def run_spin_echo(
                     returned data from previous run or `None` for the first run.
     """
     assert nEchoes == 1, "you must only choose nEchoes=1"
+    desired_p90_us = p90_us
+    prog_p90_us = prog_plen(desired_p90_us)
+    prog_p180_us = prog_plen(2*desired_p90_us)
+    print("You said you want an actual p90 of %d us and an actual p180 of %d"%(desired_p90_us,2*desired_p90_us))
+    print("So I am sending the SC a p90 time of %d and a p180 time of %d"%(prog_p90_us,prog_p180_us)
```
jmfranck, line 93: This whole added block has problems with both print statements and the intermediate variable that's not necessary. 

Stopping here

jmfranck (changes requested):

PR #30, opened 2024-02-07
Generic based echo

guinneak198:
finesse the generic ppg so that we can define the ppg list instead

### DIFF
--- a/SpinCore_pp/__init__.py
+++ b/SpinCore_pp/__init__.py
@@ -10,7 +10,8 @@ from .SpinCore_pp import (
     getData,
     stopBoard,
     tune,
-    adc_offset
+    adc_offset,
+    prog_plen
 )
 from .config_parser_fn import configuration
 from .calc_vdlist import vdlist_from_relaxivities, return_vdlist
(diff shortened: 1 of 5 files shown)


SpinCore_pp/config_parser_fn.py
```diff
@@ -182,6 +182,12 @@ class configuration(object):
             0,
             "number of echo experiments performed for a particular sample that day- usually incremented when getting on resonance",
         ),
+        "generic_echo_counter": (
+            int,
+            "file_names",
+            0,
+            "number of echo experiments performed for a particular sample that day- usually incremented when getting on resonance",
+        ),
```
guinneak198, line 190: for incrementing the name in the example

examples/generic_echo.py
```diff
@@ -0,0 +1,115 @@
...
+config_dict["generic_echo_counter"] += 1
+filename = (
+    f"{config_dict['date']}_{config_dict['chemical']}_{config_dict['type']}"
+)
+# }}}
+# {{{set phase cycling
+ph1_cyc = r_[0, 1, 2, 3]
+ph2_cyc = r_[0]
+nPhaseSteps = len(ph1_cyc) * len(ph2_cyc)
+# }}}
```
guinneak198, line 33: ph cycling used in run_hahn_echo.py

PR #34, opened 2024-06-26
new function to save and handle multiple nodes

guinneak198:


SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,70 @@
+import h5py
+import os
+from pyspecdata import *
```
jmfranck, line 3: don't do this -- do `import pyspecdata as psd`

jmfranck: actually, I just fixed by running flake and editing

SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,70 @@
...
+        config_dict pulled from the active.ini file
+    counter_type: str
+        type of counter you are incrementing
+
+    Returns
+    =======
+    config_dict: dict
+        the updated config dict after appropriately incrementing the counter
+    """
+    target_directory = getDATADIR(exp_type=my_exp_type)
```
jmfranck: actually, to keep it simple, just pass the target directory

examples/run_CPMG.py
```diff
@@ -13,15 +13,13 @@
...
 import SpinCore_pp
-from SpinCore_pp import prog_plen, get_integer_sampling_intervals
+from SpinCore_pp import prog_plen, get_integer_sampling_intervals, save_data
 from SpinCore_pp.ppg import generic
 from datetime import datetime
 from Instruments.XEPR_eth import xepr
-import h5py
 
 my_exp_type = "ODNP_NMR_comp/CPMG"
-target_directory = getDATADIR(exp_type=my_exp_type)
```
jmfranck, line 23: define the target directory here  still so that you can pass it to save_data later

SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,76 @@
...
+        config_dict pulled from the active.ini file
+    counter_type: str
+        type of counter you are incrementing
+
+    Returns
+    =======
+    config_dict: dict
+        the updated config dict after appropriately incrementing the counter
+    """
+    target_directory = psd.getDATADIR(exp_type=my_exp_type)
```
jmfranck, line 28: still not fixed

examples/run_CPMG.py
jmfranck: still not fixed

SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,76 @@
...
+    Parameters
+    ==========
+    dataset: nddata
+        acquired data in nddata format
+    my_exp_type: str
+        directory on the share drive you want to save to
+    config_dict: dict
+        config_dict pulled from the active.ini file
+    counter_type: str
+        type of counter you are incrementing
```
jmfranck, line 21: doesn't match actual arguments

SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,76 @@
...
+    =======
+    config_dict: dict
+        the updated config dict after appropriately incrementing the counter
+    """
+    # {{{ create filename
+    filename = f"{config_dict['date']}_{config_dict['chemical']}_{config_dict['type']}"
+    # }}}
+    nodename = config_dict["type"] + "_" + str(config_dict["%s_counter" % counter_type])
+    dataset.name(nodename)
+    filename_out = filename + ".h5"
```
jmfranck, line 33: doesn't make sense to define the two variables for filename here

SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,76 @@
...
+                config_dict["%s_counter" % counter_type] += 1
+                nodename = (
+                    config_dict["type"]
+                    + "_"
+                    + str(config_dict["%s_counter" % counter_type])
+                )
+            dataset.name(nodename)
+    dataset.hdf5_write(f"{filename_out}", directory=target_directory)
+    print("\n** FILE SAVED IN TARGET DIRECTORY ***\n")
+    target_directory_parts = target_directory.split("\\")
```
jmfranck, line 49: this doesn't work across operating systems.  Use `os.path.split`

SpinCore_pp/save_data.py
guinneak198: fixed

SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,76 @@
...
+    my_exp_type = "/".join((target_directory_parts[-3], target_directory_parts[-2]))
```
jmfranck, line 50: this line is really problematic, because it's not going to work in general.  I think we have two options:
* yes, because of this go back to using exp_type and converting it to a path here (*as well as* at the top of the file where you are converting to a path now)
* passing both the path and the exp type to here

SpinCore_pp/save_data.py
jmfranck: but see the comment below about exp_type, which should make this not necessary anyways

SpinCore_pp/save_data.py
jmfranck: so, to be clear, I'm stepping back what I said earlier -- I didn't know that we *needed* exp_type for this function.

By "convert to path" I mean via `getDATADIR`

I'm not walking back, however, the idea that you *do* want to check that the directory exists at the *top* of your ppg, because you don't want to run the whole thing and then get weird problems b/c the dir doesn't exist

guinneak198: I ran it and it works but I implemented a different version and again ran the examples and it works so what is there now should be good unless you still don't like the format

SpinCore_pp/save_data.py
```diff
@@ -0,0 +1,74 @@
...
+                config_dict["%s_counter" % counter_type] += 1
+                nodename = (
+                    config_dict["type"]
+                    + "_"
+                    + str(config_dict["%s_counter" % counter_type])
+                )
+            dataset.name(nodename)
+    dataset.hdf5_write(f"{filename_out}", directory=target_directory)
+    print("\n** FILE SAVED IN TARGET DIRECTORY ***\n")
+    my_exp_type = '/'.join([os.path.split(os.path.split(os.path.split(target_directory)[0])[0])[1], os.path.split(os.path.split(target_directory)[0])[1]])
```
guinneak198, line 48: takes the last two directories so this prints ODNP_NMR_comp/Echoes for example if the target directory = 'C:\\Users\\aguinnes\\Syracuse University\\Group-FranckGroup - General\\exp_data\\ODNP_NMR_comp\\Echoes\\' try it in ipython it works

examples/active.ini~
```diff
@@ -0,0 +1,46 @@
...
+uw_dip_width_ghz = 0.008
+fir_rep = 511826.9818696455
+
+[file_names]
+date = 240209
+chemical = 9p4mM_TEMPOL
+type = ODNP
+echo_counter = 17
+cpmg_counter = 3
+odnp_counter = 6
```
guinneak198, line 45: I'm not sure why this is showing up as new I diffed the active.ini against master and made sure they are the same so this didn't happen

examples/run_CPMG.py
```diff
@@ -33,22 +32,20 @@
...
 if len(sys.argv) == 2 and sys.argv[1] == "stayput":
     adjust_field = False
 # }}}
 input(
-    "I'm assuming that you've tuned your probe to",
-    config_dict["carrierFreq_MHz"],
-    "since that's what's in your .ini file. Hit enter if this is true",
+    "I'm assuming that you've tuned your probe to %f since that's what's
+    in your .ini file. Hit enter if this is true" %
+    config_dict["carrierFreq_MHz"]
```
guinneak198, line 48: input only takes one kwarg

examples/run_Hahn_echo.py
```diff
@@ -50,9 +45,8 @@
 nPhaseSteps = 4
 # }}}
 input(
-    "I'm assuming that you've tuned your probe to",
-    config_dict["carrierFreq_MHz"],
-    "since that's what's in your .ini file. Hit enter if this is true",
+    "I'm assuming that you've tuned your probe to %f since that's what's in your .ini file. Hit enter if this is true"
+    % config_dict["carrierFreq_MHz"]
```
guinneak198, line 49: again was getting error that input only accepts one kwarg

examples/run_Hahn_echo.py
```diff
@@ -101,45 +95,13 @@
...
-    [
-        "python",
-        "examples/proc_raw.py",
-        data.name(),
-        filename_out,
-        my_exp_type,
-    ],
-    cwd=r"C:\git\proc_scripts",
-    shell=True,
-)
```
guinneak198, line 145: moved to save_data function

SpinCore_pp/save_data.py
jmfranck: objectively what is there is not good

SpinCore_pp/save_data.py
jmfranck: objectively, what is there is not good.

examples/active.ini~
jmfranck: make sure you pull master too

SpinCore_pp/save_data.py
guinneak198: so are we going back to feeding the function my_exp_type instead of target_directory?

SpinCore_pp/save_data.py
guinneak198: we decided against this in other comments

SpinCore_pp/save_data.py
guinneak198: again we decided against this

SpinCore_pp/save_data.py
guinneak198: now it does

SpinCore_pp/save_data.py
guinneak198: fixed and takes my_exp_type as kwarg now

PR #35, opened 2024-06-28
WK debug run_hahn for running nutation on new probe

wfkincaid:


wfkincaid:
change to run_nutation is a similar change to the use of `input()` in run_hahn_echo, i would like to make this change match before making the full pull request. I also added my version of `find_gamma_eff.py` which I needed to get on resonance since I cant use the gui with the new probe. There were also needed changes to the active.ini to run in examples and I added some file type to .gitignore to have a less cluttered untracked files list

examples/find_gamma_eff.py
```diff
@@ -0,0 +1,14 @@
+from pylab import *
```
jmfranck, line 1: docstring -- what is this for?

PR #36, opened 2024-06-28
can run proc_raw in run_hahn_echo.py

guinneak198:
subprocess can run proc_raw immediately after acquistion

### DIFF
--- a/examples/run_Hahn_echo.py
+++ b/examples/run_Hahn_echo.py
@@ -11,6 +11,7 @@ the NMR computer to set the field etc.
 
 from pylab import *
 from pyspecdata import *
+import pyspecProcScripts
 import subprocess, os
 import SpinCore_pp
 from SpinCore_pp import get_integer_sampling_intervals
@@ -30,7 +31,7 @@ config_dict = SpinCore_pp.configuration("active.ini")
     config_dict["acq_time_ms"],
 ) = get_integer_sampling_intervals(
     SW_kHz=config_dict["SW_kHz"],
-    time_per_segment_ms=config_dict["acq_time_ms"],
+    acq_time_ms=config_dict["acq_time_ms"],
 )
 # }}}
 # {{{create filename and save to config file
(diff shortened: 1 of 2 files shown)


examples/run_Hahn_echo.py
```diff
@@ -53,7 +54,7 @@
     "I'm assuming that you've tuned your probe to",
     config_dict["carrierFreq_MHz"],
     "since that's what's in your .ini file. Hit enter if this is true",
-)
+    )
```
jmfranck, line 57: this is weird -- do you need to re-run black?

examples/run_Hahn_echo.py
```diff
@@ -132,14 +133,10 @@
...
-        "python",
-        "examples/proc_raw.py",
-        data.name(),
-        filename_out,
-        my_exp_type,
-    ],
-    cwd=r"C:\git\proc_scripts",
-    shell=True,
-)
+env = os.environ
```
jmfranck, line 136: just put this below -- dont' define intermediate variable

examples/run_Hahn_echo.py
```diff
@@ -132,14 +133,10 @@
...
+subprocess.call("which python",shell=True)
```
jmfranck, line 137: this is diagnostic -- don't do this here

examples/run_Hahn_echo.py
```diff
@@ -132,14 +133,10 @@
...
+subprocess.call((" ".join(["python",os.path.join(
+    os.path.split(os.path.split(pyspecProcScripts.__file__)[0])[0],
+    "examples", "proc_raw.py"),
+    data.name(),filename_out, my_exp_type])),
+    env=env)
```
jmfranck, line 142: but we don't want to do this here, we want to do this inside your new function.  I think fix the issues with the new function first, and then do this there.

examples/run_Hahn_echo.py~
```diff
@@ -0,0 +1,157 @@
+"""
```
jmfranck, line 1: don't commit this

jmfranck:
I think you know what to do, but it needs to go into your new function after that PR is merged.  And that PR has a bunch of simple comments that have not been dealt with yet (I assume you have the github app hooked up to slack
)

PR #38, opened 2024-06-28
new ppg for more extensive CPMG

jmfranck:


examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,224 @@
...
+        (
+            "delay",
+            config_dict["tau_us"]
+            - 2 * prog_p90_us / pi
+            - marker_us
+            - config_dict["deblank_us"],
+        ),
+        ("delay", marker_us),  # placeholder for marker
+        ("delay_TTL", config_dict["deblank_us"]),
+        ("pulse_TTL", prog_p180_us, "ph_cyc", ph3_cyc),
```
guinneak198, line 143: shouldn't this be ph2_cyc?

examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,224 @@
...
+    ret_data=None,
+    manual_echoes=1,
+)
+# }}}
+# {{{ chunk and save data
+data.chunk(
+    "t",
+    ["ph2", "ph_diff", "nEcho", "t2"],
+    [len(ph2), len(ph_diff), config_dict["nEchoes"], -1],
+)
```
guinneak198, line 194: wouldn't we want to chunk differently?

examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,224 @@
...
+data.setaxis("nEcho", r_[0 : config_dict["nEchoes"]]).setaxis(
+    "ph2", ph2 / 4
```
guinneak198, line 196: this should be config_dict['nEchoes']+1

examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,224 @@
...
+).setaxis("ph_diff", ph_diff / 4)
+data.set_prop("postproc_type", "spincore_diffph_SE_v2")
+data.set_prop("coherence_pathway", {"ph_overall": -1, "ph1": +1})
```
guinneak198, line 199: new coherence pathway needed

examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,224 @@
...
+# {{{ chunk and save data
+data.chunk(
+    "t",
+    ["ph2", "ph_diff", "nEcho", "t2"],
+    [len(ph2), len(ph_diff), config_dict["nEchoes"], -1],
+)
+data.setaxis("nEcho", r_[0 : config_dict["nEchoes"]]).setaxis(
+    "ph2", ph2 / 4
+).setaxis("ph_diff", ph_diff / 4)
+data.set_prop("postproc_type", "spincore_diffph_SE_v2")
```
guinneak198, line 198: new postproc_type needed

examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,224 @@
...
+            data.name(nodename)
+            config_dict["cpmg_counter"] += 1
+data.hdf5_write(f"{filename_out}", directory=target_directory)
+print("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
+print(
+    "saved data to (node, file, exp_type):",
+    data.name(),
+    filename_out,
+    my_exp_type,
+)
```
guinneak198, line 222: this can all be replaced with save_data

examples/run_largecyc_CPMG.py
jmfranck: yes correct, please change

examples/run_largecyc_CPMG.py
jmfranck: YES! Please do fix -- the first 3 dimensions should be ph1 ph2 ph_overall

examples/run_largecyc_CPMG.py
wfkincaid: ph2_cyc also makes more sense to me comparing with planned pulse program

examples/run_largecyc_CPMG.py
jmfranck: no -- I subtract 1 from the loop so that there are still nEchoes

examples/run_largecyc_CPMG.py
jmfranck: yes, let's just go with:
```
{'ph1': +1,
'ph2": -2,
'ph_overall':-1,
}
```
note that this will only pull the odd echoes, but we have  to pick something for now.

(we can make this more complicated by using "eval", but that's beyond this PR)

examples/run_largecyc_CPMG.py
jmfranck: please do!!!

SpinCore_pp/ppg/generic.py
```diff
@@ -32,6 +31,7 @@ def generic(
     ph2_cyc=r_[0],
     ret_data=None,
     amplitude=1.0,
```
wfkincaid, line 33: do we want `amplitude=1.0` if we may soon use this with the new probe? Why not just include amplitude as an argument without presetting it to 1?

wfkincaid: amplitude is also not included in the description of each parameter below

examples/run_largecyc_CPMG.py
guinneak198: this is a PR for laod_Data.py in proc_scripts

wfkincaid: to confirm, the suggestion is to add to load_data.py something like
`def proc_spincore_generalproc_v1(s, fl=None):
    for j in [k for k in s.dimlabels if k.startswith('ph')]:
        s.ft(j)`
for a lookup called `spincore_generalproc_v1`

wfkincaid: this should then be a small PR in `proc_scripts/pyspecProcScripts`

examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,203 @@
...
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["cpmg_counter"] += 1
+# }}}
+# {{{ command-line option to leave the field untouched (if you set it once, why set it again)
+adjust_field = True
+if len(sys.argv) == 2 and sys.argv[1] == "stayput":
+    adjust_field = False
+# }}}
+input(
+    "I'm assuming that you've tuned your probe to %f since that's what's
```
jmfranck, line 59: you can't linebreak in the middle of a string like this -- that's why flake is failing right now

examples/run_largecyc_CPMG.py
```diff
@@ -0,0 +1,203 @@
...
+    SW_kHz=config_dict["SW_kHz"],
+    ret_data=None,
+    manual_echoes=1,
+)
+# }}}
+# {{{ chunk and save data
+data.chunk(
+    "t",
+    ["ph1","ph2","ph_overall", "nEcho", "t2"],
+    [len(ph2), len(ph_diff), config_dict["nEchoes"], -1],
```
jmfranck, line 189: line 189 here doesn't match line 188

examples/run_largecyc_CPMG.py
jmfranck: note that the second line here needs to change correspondingly (I think I have another comment about this)

SpinCore_pp/ppg/generic.py
jmfranck: when you define a keyword argument, you have to give it a default.

examples/run_largecyc_CPMG.py
jmfranck: (there are / should be other flake errors after this, but b/c this one is breaking, that's all it's giving you until you fix it --> you can see the flake errors by clicking for more info where it tells you it fails.)

examples/run_largecyc_CPMG.py
jmfranck: (meaning that you need to give the length of the phase cycles you refer to on the previous line)

guinneak198 (commented):
changed to import the functions use especially since pyspecdata is only called for getDATADIR otherwise looks good

PR #39, opened 2024-07-01
fix typo in run_nutation

guinneak198:


examples/run_nutation.py
```diff
@@ -79,20 +80,18 @@
...
-nutation_data.set_units("t2", "s")
-nutation_data.set_prop("postproc_type", "spincore_nutation_v4")
-nutation_data.set_prop("coherence_pathway", {"ph1": +1})
-nutation_data.set_prop("acq_params", config_dict.asdict())
+    data.setaxis("nScans", r_[0 : config_dict["nScans"]])
+data.reorder(["ph1", "nScans", "t2"])
+data.set_units("t2", "s")
+data.set_prop("postproc_type", "spincore_nutation_v4")
+data.set_prop("coherence_pathway", {"ph1": +1})
+data.set_prop("acq_params", config_dict.asdict())
```
jmfranck, line 95: while it's good to try to limit the diff, I notice that we keep having problems with ppgs, where the `data` variable is named differently, so I fix that.

PR #41, opened 2024-07-02
Fix large cpmg bug

jmfranck:


guinneak198 (approved):
yes!

examples/run_largecyc_CPMG.py
```diff
@@ -190,7 +190,7 @@
 )
 data.setaxis("nEcho", r_[0 : config_dict["nEchoes"]]).setaxis(
     "ph1", ph1 / 4
-).setaxis("ph2", ph2).setaxis("ph_overall", ph_overall / 4)
+).setaxis("ph2", ph2 / 4).setaxis("ph_overall", ph_overall / 4)
```
jmfranck, line 193: note that this will throw off the amplitude of the signal that comes out of the ft (without division, signal will be a factor of 4 larger than it should be) it also messes up the axis, though we don't actually use the axis coordinates much.

PR #42, opened 2024-07-03
store the programmed pulse lengths, so that they are saved.

jmfranck:


guinneak198 (commented):
This looks okay but I am concerned how we want to handle other ppg functions - the prog_plen function is called inside the run_spin_echo function vs inside the run_hahn_echo.py where the config file is saved, should we just not worry about that at this stage?

jmfranck:
In the same way.

PR #44, opened 2024-07-08
Adding amplitude param

wfkincaid:


wfkincaid:
other than the use of `except Exception:` in run_field_dep_justMw.py and combined_ODNP.py, i think everything else is all set for this

examples/combined_ODNP.py
```diff
@@ -150,6 +167,7 @@
         filename = "temp_ctrl.h5"
         DNP_data.hdf5_write(filename, directory=target_directory)
         final_log.append("change the name accordingly once this is done running!")
+    raise
```
jmfranck, line 170: This causes it to quit

jmfranck (changes requested):

examples/combined_ODNP.py
wfkincaid: so we don't want `raise` since if there is an error, this will be thrown forcing the script to stop running. Is what I did with `except Exception:` okay?

examples/combined_ODNP.py
```diff
@@ -9,10 +9,12 @@
 """
 import numpy as np
 from numpy import r_
-from pyspecdata import *
+import pyspecdata as psd
 from pyspecdata.file_saving.hdf_save_dict_to_group import hdf_save_dict_to_group
-from pyspecdata import strm
-import os, sys, time
+from pyspecdata import strm, DNP_data
```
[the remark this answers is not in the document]
wfkincaid: I don't think this would even work. I though DNP_data was something in pyspecdata from the flake assertions but I'm realizing part of my confusion was where the variable is initialized vs where its first used

jmfranck:
I'm going to close this, and create a new joined repo

PR #45, opened 2024-07-13
Better nutation

guinneak198:
https://jmfrancklab.slack.com/archives/C06LNFNP2HM/p1720757389631039


SpinCore_pp/ppg/echo.py
```diff
@@ -96,13 +96,14 @@ def run_spin_echo(
     tx_phases = r_[0.0, 90.0, 180.0, 270.0]
     nPhaseSteps = len(ph1_cyc) * len(ph2_cyc)
     data_length = 2 * nPoints * nEchoes * nPhaseSteps
+    for_RX_nScans = 1
```
guinneak198, line 99: nScans still needs to be it's own variable but this this prevents having to multiply by nScans in the post processing function.

SpinCore_pp/ppg/echo.py
```diff
@@ -96,13 +96,14 @@ def run_spin_echo(
...
     for nScans_idx in range(nScans):
         run_scans_time_list = [time.time()]
         run_scans_names = ["configure"]
         configureTX(adcOffset, carrierFreq_MHz, tx_phases, amplitude, nPoints)
         run_scans_time_list.append(time.time())
         run_scans_names.append("configure Rx")
-        acq_time_ms = configureRX(SW_kHz, nPoints, nScans, nEchoes, nPhaseSteps)
+        acq_time_ms = configureRX(SW_kHz, nPoints, for_RX_nScans, nEchoes, nPhaseSteps)
```
guinneak198, line 106: I'm pretty sure configureRX is where we want to feed one scan but we might need to remove the `acq_time_ms = `  since this value is used in the actual ppg (line 121) and will overwrite the acq_time we want to use?

examples/run_nutation.py
```diff
@@ -25,18 +25,17 @@
     nPoints,
     config_dict["SW_kHz"],
     config_dict["acq_time_ms"],
-) = get_integer_sampling_intervals(
-    config_dict["SW_kHz"], config_dict["acq_time_ms"]
-)
+) = get_integer_sampling_intervals(config_dict["SW_kHz"], config_dict["acq_time_ms"])
```
guinneak198, line 28: black

examples/run_nutation.py
```diff
@@ -82,16 +82,17 @@
...
+data.chunk("t", ["ph2", "ph1", "t2"], [2, 2, -1])
+data.setaxis("ph1", ph1_cyc / 4).setaxis("ph2", ph2_cyc / 4)
 if config_dict["nScans"] > 1:
     data.setaxis("nScans", r_[0 : config_dict["nScans"]])
-data.reorder(["ph1", "nScans", "t2"])
+data.reorder(["nScans", "ph2", "ph1", "p_90", "t2"])
 data.set_units("t2", "s")
-data.set_prop("postproc_type", "spincore_nutation_v4")
-data.set_prop("coherence_pathway", {"ph1": +1})
+data.set_prop("postproc_type", "spincore_nutation_v3")
```
guinneak198, line 94: I looked through the load_data and there is already a postprocessing function that suits this phase cycling!

examples/run_nutation.py
```diff
@@ -82,16 +82,17 @@
...
+data.setaxis("p_90", p90_range_us * 1e-6).set_units("p_90", "s")
 # {{{ chunk and save data
-data.chunk("t", ["ph1", "t2"], [4, -1])
-data.setaxis("ph1", ph1_cyc / 4)
+data.chunk("t", ["ph2", "ph1", "t2"], [2, 2, -1])
+data.setaxis("ph1", ph1_cyc / 4).setaxis("ph2", ph2_cyc / 4)
 if config_dict["nScans"] > 1:
     data.setaxis("nScans", r_[0 : config_dict["nScans"]])
-data.reorder(["ph1", "nScans", "t2"])
+data.reorder(["nScans", "ph2", "ph1", "p_90", "t2"])
```
guinneak198, line 92: move nScans to outside and 90 times to inside

SpinCore_pp/ppg/echo.py
jmfranck: Yes!

examples/run_nutation.py
```diff
@@ -82,16 +82,17 @@
         SW_kHz=config_dict["SW_kHz"],
         ret_data=data,
     )
-data.setaxis("indirect", p90_range_us * 1e-6).set_units("indirect", "s")
+data.rename("indirect", "p_90")
+data.setaxis("p_90", p90_range_us * 1e-6).set_units("p_90", "s")
 # {{{ chunk and save data
-data.chunk("t", ["ph1", "t2"], [4, -1])
-data.setaxis("ph1", ph1_cyc / 4)
+data.chunk("t", ["ph2", "ph1", "t2"], [2, 2, -1])
```
jmfranck, line 88: Is this correct? Do you list the phase cycle in reverse order?  I am asking because I do not know.... Would need to look at another example and I'm doing this on my phone

examples/run_nutation.py
jmfranck: I would call it 5, I think.

Part of the function of this is to label the version number of the ppg, and we're not sure that this does the same thing as v3

In the more recent versions of procscripts, I have the generic post-processing function which should handle most cases....  So we would just end up adding a line to the dictionary

jmfranck (changes requested):

SpinCore_pp/ppg/echo.py
jmfranck: please `grep -Rile 'run_spin_echo'` in the examples directory to find the programs that use this function, and then just increment the version number of the post proc by 1 for all of those files. 

This way we know that they were acquired with a version that does not make us multiply by n scans in post-processing

examples/run_nutation.py
guinneak198: I am basing this off of how the IR is chunked in combined_ODNP which uses ph1_cyc and ph2_cyc and seems to be chunked correctly

SpinCore_pp/ppg/echo.py
guinneak198: I incremented versions and ran into issues needing to do some cleanup for them to pass flake but should be good now

SpinCore_pp/run_nutation_amplitude.py
```diff
@@ -1,108 +1,118 @@
-from pyspecdata import *
-import os,sys,time
+from pyspecdata import figlist_var, getDATADIR, r_, ndshape, concat
+import os
```
guinneak198, line 2: to pass flake

SpinCore_pp/run_nutation_amplitude.py
```diff
@@ -1,108 +1,118 @@
...
 import SpinCore_pp
 from datetime import datetime
-from pyspecdata.file_saving.hdf_save_dict_to_group import hdf_save_dict_to_group
```
guinneak198, line 5: unused

SpinCore_pp/run_nutation_amplitude.py
```diff
@@ -1,108 +1,118 @@
...
-config_dict['echo_counter'] += 1
+# {{{importing acquisition parameters
+config_dict = SpinCore_pp.configuration("active.ini")
+nPoints = int(config_dict["acq_time_ms"] * config_dict["SW_kHz"] + 0.5)
+# }}}
+# {{{create filename and save to config file
+date = datetime.now().strftime("%y%m%d")
+config_dict["type"] = "nutation"
+config_dict["date"] = date
+config_dict["echo_counter"] += 1
```
guinneak198, line 24: black

SpinCore_pp/run_nutation_amplitude.py
```diff
@@ -1,108 +1,118 @@
...
+print(
+    "I'm assuming that you've tuned your probe to",
+    config_dict["carrierFreq_MHz"],
+    "since that's what's in your .ini file",
+)
+Field = config_dict["carrierFreq_MHz"] / config_dict["gamma_eff_MHz_G"]
+print(
+    "Based on that, and the gamma_eff_MHz_G you have in your .ini file, I'm setting the field to %f"
+    % Field
+)
```
guinneak198, line 37: black

SpinCore_pp/run_nutation_amplitude.py
```diff
@@ -1,108 +1,118 @@
...
     assert Field > 3300, "are you crazy?? field is too low!"
     Field = x.set_field(Field)
-    print("field set to ",Field)
-#}}}
-#{{{ phase cycling
-tx_phases = r_[0.0,90.0,180.0,270.0]
+    print("field set to ", Field)
+# }}}
+# {{{ phase cycling
+tx_phases = r_[0.0, 90.0, 180.0, 270.0]
```
guinneak198, line 45: black

SpinCore_pp/run_nutation_amplitude.py
```diff
@@ -1,108 +1,118 @@
...
 phase_cycling = True
 if phase_cycling:
-    ph1 = r_[0,1,2,3]
-    ph2 = r_[0,2]
+    ph1 = r_[0, 1, 2, 3]
+    ph2 = r_[0, 2]
     nPhaseSteps = 8
```
guinneak198, line 50: black

examples/combined_ODNP.py
```diff
@@ -168,10 +166,8 @@
         os.remove("temp_ctrl.h5")
         target_directory = os.path.getcwd()
         filename = "temp_ctrl.h5"
-        DNP_data.hdf5_write(filename, directory=target_directory)
-        final_log.append(
-            "change the name accordingly once this is done running!"
-        )
+        control_thermal.hdf5_write(filename, directory=target_directory)
```
jmfranck, line 169: this change was already made in FLinst.

jmfranck:
I'm closing this.  Just diff files against FLinst to create a branch there.  I have run black on the master of FLinst, so that shouldn't be too bad.

